PR #4, opened 2023-11-27
RFC-0003: Asynchronous Circuits in `o1js`

teddyjfpender:

Here is RFC-0003! It's about bringing asynchronous circuits to `o1js`. This enhancement is set to significantly improve zkApp design patterns and eliminate work-arounds for performing asynchronous operations with circuits.

# 🔑 Why This Change Is Crucial

Currently, `o1js` circuits are constrained by sequential execution and are unable to perform asynchronous operations, which limits their ability to interact with dynamic data sources (such as other circuits!). By adding asynchronous circuit capabilities, we can overcome these limitations, offering a boost to zkApp design flexibility.

# 🔍 Exploring Practical Use Cases

There are a few detailed scenarios like dynamic data integration in smart contract circuits and real-time player attribute updates in games that touch on a small part of the large impact this feature can have across a variety of applications.

# 💭 Seeking Your Input

Your expertise, implementation designs, feedback, and creative use-case scenarios are invaluable in shaping a robust implementation.

# 📖 For More Details

Please refer to the full RFC document for an in-depth understanding of the scope, potential impact, and the challenges addressed.

# 🤝 Join the Discussion

This is an open invitation to all developers, users, and stakeholders within the Mina ecosystem to be a part of this exciting development. Let's work together to take the Mina ecosystem to new heights of innovation and functionality!

### DIFF
--- a/RFCs/rfc-0003-async-circuit-witnesses.md
+++ b/RFCs/rfc-0003-async-circuit-witnesses.md
@@ -1,0 +1,59 @@
+# RFC-0003: Asynchronous Circuits in `o1js`
+
+- **Intent**: To introduce asynchronous circuit in `o1js`, allowing for improved zkApp design patterns and improved program concurrency in circuit development.
+- **Submitted by**: Theodore Pender (Github: @teddyjfpender, email: theodore.pender@minaprotocol.com, Twitter/X: @franklyteddy)
+- **Submitted on**: Monday, November 27, 2023
+
+## Abstract
+
+This RFC proposes the implementation of asynchronous circuits and thus asynchronous circuit witnesses in `o1js`, addressing the limitations of sequential execution in zkApp development. The proposed feature enables circuits to pause and await data fetched asynchronously, thus enhancing the richness of zkApp design patterns and improving the efficiency and concurrency of program execution.
+
+## Introduction
+
+Circuits built in `o1js` currently cannot await the result of asynchronous computations and are constrained by sequential execution, limiting the ability to integrate dynamic data not known at the start of execution. This RFC proposes adding  asynchronous circuit as a solution, permitting circuits to pause their execution and resume upon receiving necessary data, be it from API calls, other circuits, or other sources of data.
+
+Problem: Circuits cannot await the result of asynchronous computations and therefore `o1js` users must find solutions outside that are potentially error-prone and hinder brilliant developer experiences. 
+Solution: Implementing asynchronous circuits in `o1js` to allow circuits to pause and resume, integrating data from external sources.
+
+## Objectives
+
+- Demonstrate how this feature enhances program concurrency and expands zkApp design possibilities.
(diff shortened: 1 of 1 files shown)


halsaphi (approved):

mitschabaude:
Quick note on implementation:

An o1js circuit is given as a callback to Pickles, which gives it as a callback to Snarky at various steps of compiling/proving. In both Pickles and Snarky, the circuit is expected to finish synchronously. Therefore, the main lift is to adapt both of those libraries to allow async circuits.

We already had to allow dealing with JS Promises in other places of Pickles (like when it waits for a Kimchi proof to finish), therefore there is an OCaml library called `Promise` ([code here](https://github.com/MinaProtocol/mina/tree/develop/src/lib/promise)) which implements Promises for both native OCaml and JS. The goal would be to make a circuit coming from o1js return a `Promise.t`, and thread that through Pickles.

A good starting point to enable async circuit is probably the `checked_runner.ml` module of snarky, particularly the `type run` which describes how a function `'t` is run to produce a value `'a`, and in the process update a state `run_state`:

https://github.com/o1-labs/snarky/blob/94b2df82129658d505b612806a5804bc192f13f0/src/base/checked_runner.ml#L433

```ocaml
// checked_runner.ml
type ('a, 't) run = 't -> run_state -> run_state * 'a
```

In the new model, we instead want that type to return `(run_state * 'a) Promise.t`. An interesting exercise (which could help you get to know snarky) is to just change the type and then start fixing everything that breaks:
```diff
// checked_runner.ml
- type ('a, 't) run = 't -> run_state -> run_state * 'a
+ type ('a, 't) run = 't -> run_state -> (run_state * 'a) Promise.t
```

(To make this work, you need to add the promise library to the `libraries` to snarky's [dune file](https://github.com/o1-labs/snarky/blob/94b2df82129658d505b612806a5804bc192f13f0/src/base/dune#L5-L6) 
```diff
// snarky/src/base/dune
 (libraries bitstring_lib core_kernel h_list interval_union snarky_intf promise
-   bignum.bigint)
+ promise bignum.bigint)
```

For dealing with promises, you could use monadic let syntax, like this:
```diff
// example code using `run`
- let state, result =  run circuit state in
+ let%map.Promise state, result =  run circuit state in
```
which magically will make the function this is done in return a `Promise.t`.

However, instead of literally changing the `run` type and fixing everything that breaks, a smoother (and recommended) approaches could to introduce a sibling type, `run_async`, and progressively use that in more and more places:

```diff
// checked_runner.ml
type ('a, 't) run = 't -> run_state -> run_state * 'a
+ type ('a, 't) run_async = 't -> run_state -> (run_state * 'a) Promise.t
```

This would be my preferred approach personally!

Matthew also suggested functoring over the `run` type entirely, so different components of the system could make it be either async or sync.

mitschabaude:
Implementation is under way! https://github.com/o1-labs/o1js/pull/1450

PR #7, opened 2023-12-07
RFC-0004: Decreasing the minimum interval between SmartContract's Transactions from 3 minutes to 1 second with Recursive Proofs

dfstio:
This RFC aims to improve the Mina Protocol's developer and user experience by enabling faster transaction processing in SmartContracts. Currently, a minimum interval of 3 minutes is required between SmartContract transactions that have preconditions and alter the state. This proposal intends to shorten this interval to 1 second, allowing rapid transaction execution with state preconditioning and changing.

### DIFF
--- a/RFCs/rfc-0004-fast-transactions.md
+++ b/RFCs/rfc-0004-fast-transactions.md
@@ -1,0 +1,77 @@
+# RFC-0004: Enhancing SmartContract Transactions with Recursive Proofs
+
+- **Intent**: Implement recursive proofs in SmartContract transactions to reduce the minimum transaction interval from 3 minutes to 1 second.
+- **Submitted by**: DFST ([Github](https://github.com/dfstio), [Twitter](https://twitter.com/dfst_io), Discord: dfst.io)
+- **Date**: Tuesday, December 05, 2023
+
+## Abstract
+
+This RFC aims to improve the Mina Protocol's developer and user experience by enabling faster transaction processing in SmartContracts. Currently, a minimum interval of 3 minutes is required between SmartContract transactions that have preconditions and alter the state. This proposal intends to shorten this interval to 1 second, allowing rapid transaction execution with state preconditioning and changing.
+
+## Introduction
+
+Mina Protocol's SmartContract transactions, authenticated through proof, must meet these criteria:
+
+- Matching account update preconditions with the current account state.
+- Proof verification against the SmartContract's verification key.
+- Fee alignment with the market rate.
+
+Once verified, the transaction's inclusion in the block is assured, bypassing the need for blockchain-wide recalculations. Despite known post-verification states, SmartContracts must wait about 3 minutes for state reflection in the new block before initiating another transaction.
+
+Current approaches to multiple transactions within a block include:
+
+- Non-use of preconditions and nonce manipulation (for example, in case of the stateless contract).
+- Recursive proofs via ZkProgram for bundled transactions.
+- Using Actions and Reducer
+
+These methods, especially recursive proofs, pose significant entry barriers due to complexity and infrastructure demands and do not allow for many uses as in the case with Actions and Reducer.
+
+**Problem**: The Mina blockchain's slow transaction pace (3 to 20 minutes) hinders rapid processing, contrasting with other blockchains like Tron (3 seconds) and Ethereum (12 seconds).
+
+**Relevance**: Accelerating transaction speed can greatly enhance the blockchain's Total Value Locked (TVL) and user experience, fostering Mina's ecosystem growth.
+
+## Objectives
+
+- Reduce SmartContract transaction intervals from 3 minutes to 1 second.
+- Introduce and utilize the concept of "guaranteed new state".
+- Enable immediate use of the new guaranteed state post-verification.
+- Simplify SmartContract interaction with the new state.
+- Integrate the changes to the new light web node to receive new guaranteed state within 50-100 ms.
+
+## Motivation and Rationale
+
+The need for rapid transaction processing is driven by:
+
+- Current system limitations: Transactions can take 3 to 20 minutes.
+- Real-world applications: Many scenarios require transaction intervals of only a few seconds.
+- Ecosystem growth: Faster transactions open up more use cases and attract developers.
+
+## Possible Implementations
+
+### Variant 1: Node-side Recursive Proof Computation
+
+Node, after transaction verification, compute the new account state, sign it, and return it to the zkApp. The zkApp can utilize this state (for example, by providing it as an argument to the fetchAccount()) to create new immediate transactions. The node can replace the existing mempool transaction with a new composite one, reflecting the initial transaction's preconditions and the final state of the last transaction.
+
+### Variant 2: zkApp-side Recursive Proof Computation
+
+SmartContracts gain the ability to merge multiple transaction proofs. The first transaction proceeds normally, with subsequent transactions are based on the internal state maintained by o1js and calculation of the merged proofs. The composite transaction, combining initial preconditions, merged proof and final state, is sent for replacement to the node.
+
+## Scenarios and Use Cases
+
+### Scenario 1: High-Frequency Transaction SmartContract
+
+Example: A chess blitz game SmartContract needing several transactions per minute. This feature would enable high-frequency transactions without complex proof infrastructure.
+
+### Scenario 2: Multi-User SmartContract Requiring Immediate Final State
+
+Example: A multi-user SmartContract that operates on simultaneous requests and requires immediate access to the final state post-transaction. It can use Actions and Reducer to collect user requests, then send reduce transaction and receive the final state to be used in the following user's transactions immediately.
+
+## Open Issues and Discussion Points
+
+- The calculation of any proof takes some time, and the previous transaction can be included in the block at that time, so the replacement transaction will fail. There will be needed a mechanism for automatic increase of the nonce in this case to be able to send this transaction as new or other ways how to handle this situation.
+- Although the good way to use a new signed state is to feed it as an argument to fetchAccount(), the other ways that would allow to use of this feature without any changes to the existing codebase, should be considered
+- This addition, in case of the implementation in accordance with variant 1, requires changes to the node code and should be tested extensively on the testnet. The timing of this change should be coordinated with the mainnet launch, with the mainnet launch being the obvious priority to distribute engineering and testing resources wisely.
+
+## Conclusion
+
+Fast transaction capabilities in Mina protocol and `o1js` mark a significant advancement in developer and user experience, potentially catalyzing the adoption and growth of the Mina ecosystem and can greatly enhance the blockchain's Total Value Locked (TVL).
--- a/RFCs/rfc-0004-fast-transactions.md
+++ b/RFCs/rfc-0004-fast-transactions.md
@@ -2,7 +2,7 @@
 
 - **Intent**: Implement recursive proofs in SmartContract transactions to reduce the minimum transaction interval from 3 minutes to 1 second.
 - **Submitted by**: DFST ([Github](https://github.com/dfstio), [Twitter](https://twitter.com/dfst_io), Discord: dfst.io)
-- **Date**: Tuesday, December 05, 2023
+- **Date**: Thursday, December 07, 2023
 
 ## Abstract
 
@@ -16,15 +16,15 @@ Mina Protocol's SmartContract transactions, authenticated through proof, must me
 - Proof verification against the SmartContract's verification key.
 - Fee alignment with the market rate.
 
-Once verified, the transaction's inclusion in the block is assured, bypassing the need for blockchain-wide recalculations. Despite known post-verification state, SmartContracts must wait about 3 minutes for state reflection in the new block before initiating another transaction.
+Once verified, the transaction's inclusion in the block is assured, bypassing the need for blockchain-wide recalculations. Despite a known post-verification state, SmartContracts must wait about 3 minutes for state reflection in the new block before initiating another transaction.
 
 Current approaches to multiple transactions within a block include:
 
-- Non-use of preconditions and nonce manipulation (for example, in case of the stateless contract).
+- Non-use of preconditions and nonce manipulation (for example, in the case of the stateless contract).
 - Recursive proofs via ZkProgram for bundled transactions.
 - Using Actions and Reducer
 
-These methods, especially recursive proofs, pose significant entry barriers due to complexity and infrastructure demands and do not allow for many uses as in the case with Actions and Reducer.
+These methods, especially recursive proofs, pose significant entry barriers due to complexity and infrastructure demands and do not allow for many uses, as in the case of Actions and Reducer.
 
 **Problem**: The Mina blockchain's slow transaction pace (3 to 20 minutes) hinders rapid processing, contrasting with other blockchains like Tron (3 seconds) and Ethereum (12 seconds).
 
@@ -33,10 +33,10 @@ These methods, especially recursive proofs, pose significant entry barriers due 
 ## Objectives
 
 - Reduce SmartContract transaction intervals from 3 minutes to 1 second.
-- Introduce and utilize the concept of "guaranteed new state".
+- Introduce and utilize the concept of "a guaranteed new state".
 - Enable immediate use of the new guaranteed state post-verification.
 - Simplify SmartContract interaction with the new state.
-- Integrate the changes to the new light web node to receive new guaranteed state within 50-100 ms.
+- Integrate the changes to the new light web node to receive a new guaranteed state within 50-100 ms.
 
 ## Motivation and Rationale
 
@@ -50,11 +50,11 @@ The need for rapid transaction processing is driven by:
 
 ### Variant 1: Node-side Recursive Proof Computation
 
-Node, after transaction verification, compute the new account state, sign it, and return it to the zkApp. The zkApp can utilize this state (for example, by providing it as an argument to the fetchAccount()) to create new immediate transactions. The node can replace the existing mempool transaction with a new composite one, reflecting the initial transaction's preconditions and the final state of the last transaction.
+Node, after transaction verification, computes the new account state, sign it, and returns it to the zkApp. The zkApp can utilize this state (for example, by providing it as an argument to the fetchAccount()) to create new immediate transactions. The node can replace the existing mempool transaction with a new composite one, reflecting the initial transaction's preconditions and the final state of the last transaction.
 
 ### Variant 2: zkApp-side Recursive Proof Computation
 
-SmartContracts gain the ability to merge multiple transaction proofs. The first transaction proceeds normally, with subsequent transactions are based on the internal state maintained by o1js and calculation of the merged proofs. The composite transaction, combining initial preconditions, merged proof and final state, is sent for replacement to the node.
+SmartContracts gain the ability to merge multiple transaction proofs. The first transaction proceeds normally, with subsequent transactions being based on the internal state maintained by o1js and the calculation of the merged proofs. The composite transaction, combining initial preconditions, merged proof, and final state, is sent for replacement to the node.
 
 ## Scenarios and Use Cases
 
@@ -64,12 +64,12 @@ Example: A chess blitz game SmartContract needing several transactions per minut
 
 ### Scenario 2: Multi-User SmartContract Requiring Immediate Final State
 
-Example: A multi-user SmartContract that operates on simultaneous requests and requires immediate access to the final state post-transaction. It can use Actions and Reducer to collect user requests, then send reduce transaction and receive the final state to be used in the following user's transactions immediately.
+Example: A multi-user SmartContract that operates on simultaneous requests and requires immediate access to the final state post-transaction. It can use Actions and Reducer to collect user requests, then send a reduce transaction and receive the final state to be used in the following user's transactions immediately.
 
 ## Open Issues and Discussion Points
 
 - The calculation of any proof takes some time, and the previous transaction can be included in the block at that time, so the replacement transaction will fail. There will be needed a mechanism for automatic increase of the nonce in this case to be able to send this transaction as new or other ways how to handle this situation.
-- Although the good way to use a new signed state is to feed it as an argument to fetchAccount(), the other ways that would allow to use of this feature without any changes to the existing codebase, should be considered
+- Although the good way to use a new signed state is to feed it as an argument to fetchAccount(), the other ways that would allow use of this feature without any changes to the existing codebase should be considered
 - This addition, in case of the implementation in accordance with variant 1, requires changes to the node code and should be tested extensively on the testnet. The timing of this change should be coordinated with the mainnet launch, with the mainnet launch being the obvious priority to distribute engineering and testing resources wisely.
 
 ## Conclusion
--- a/RFCs/rfc-0004-fast-transactions.md
+++ b/RFCs/rfc-0004-fast-transactions.md
@@ -1,44 +1,34 @@
-# RFC-0004: Decreasing the minimum interval between SmartContract's Transactions from 3 minutes to 1 second with Recursive Proofs
+# RFC-0004: Decreasing the minimum interval between SmartContract's Transactions from 3 minutes to 1 second by providing a fast, universal L2 ephermal blockspace with no dedicated infrastructure
 
-- **Intent**: Implement recursive proofs in SmartContract transactions to reduce the minimum transaction interval from 3 minutes to 1 second.
+- **Intent**: Implement fast, universal L2 ephermal blockspace that requires no dedicated infrastructure to reduce the minimum transaction interval from 3 minutes to 1 second.
 - **Submitted by**: DFST ([Github](https://github.com/dfstio), [Twitter](https://twitter.com/dfst_io), Discord: dfst.io)
-- **Date**: Thursday, December 07, 2023
+- **Date**: Thursday, December 07, 2023 as amended on Tuesday, December 12, 2023
+- **Requires**: Changes to the o1js library and Mina protocol node
+
+![Second image](https://docs.minanft.io/rfc4/second.png)
 
 ## Abstract
 
-This RFC aims to improve the Mina Protocol's developer and user experience by enabling faster transaction processing in SmartContracts. Currently, a minimum interval of 3 minutes is required between SmartContract transactions that have preconditions and alter the state. This proposal intends to shorten this interval to 1 second, allowing rapid transaction execution with state preconditioning and changing.
+The speed and cost of sending the transactions to the blockchain are of utmost importance to end users, organizations, and developers and significantly impact the adoption of the blockchain technology and Total Value Locked in the blockchain.
 
-## Introduction
+This RFC aims to provide each zkApp a personal, fast, and universal on-demand L2 blockspace (ephermeral blockchain) with no dedicated infrastructure to improve the Mina Protocol's developer and user experience by enabling fast transaction processing in SmartContracts.
 
-Mina Protocol's SmartContract transactions, authenticated through proof, must meet these criteria:
+Currently, a minimum interval of 3 minutes is required between SmartContract transactions that use preconditions and alter the state, contrasting with other blockchains like Tron (3 seconds) and Ethereum (12 seconds). This proposal intends to shorten this interval to 1 second, allowing rapid transaction execution with state preconditioning and changing.
 
-- Matching account update preconditions with the current account state.
-- Proof verification against the SmartContract's verification key.
-- Fee alignment with the market rate.
+The focus of this proposal is to provide speed capabilities equal to the usual L2 but without the need to significantly change the code or do some integrations with third-party dedicated L2 infrastructure by describing a light ephermal L2 blockspace with a life span of about 3 minutes that, at the end of its life, as a new L1 block comes, transfers all the assets and state to the L1 and disappears to be born again when needed by zkApp.
 
-Once verified, the transaction's inclusion in the block is assured, bypassing the need for blockchain-wide recalculations. Despite a known post-verification state, SmartContracts must wait about 3 minutes for state reflection in the new block before initiating another transaction.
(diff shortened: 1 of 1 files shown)


RFCs/rfc-0004-fast-transactions.md
```diff
@@ -0,0 +1,77 @@
...
+These methods, especially recursive proofs, pose significant entry barriers due to complexity and infrastructure demands and do not allow for many uses, as in the case of Actions and Reducer.
+
+**Problem**: The Mina blockchain's slow transaction pace (3 to 20 minutes) hinders rapid processing, contrasting with other blockchains like Tron (3 seconds) and Ethereum (12 seconds).
+
+**Relevance**: Accelerating transaction speed can greatly enhance the blockchain's Total Value Locked (TVL) and user experience, fostering Mina's ecosystem growth.
+
+## Objectives
+
+- Reduce SmartContract transaction intervals from 3 minutes to 1 second.
+- Introduce and utilize the concept of "a guaranteed new state".
```
teddyjfpender, line 36: What is "a guaranteed new state"? Is that not equivalent to finality?

RFCs/rfc-0004-fast-transactions.md
```diff
@@ -0,0 +1,77 @@
+# RFC-0004: Decreasing the minimum interval between SmartContract's Transactions from 3 minutes to 1 second with Recursive Proofs
+
+- **Intent**: Implement recursive proofs in SmartContract transactions to reduce the minimum transaction interval from 3 minutes to 1 second.
+- **Submitted by**: DFST ([Github](https://github.com/dfstio), [Twitter](https://twitter.com/dfst_io), Discord: dfst.io)
+- **Date**: Thursday, December 07, 2023
+
+## Abstract
+
+This RFC aims to improve the Mina Protocol's developer and user experience by enabling faster transaction processing in SmartContracts. Currently, a minimum interval of 3 minutes is required between SmartContract transactions that have preconditions and alter the state. This proposal intends to shorten this interval to 1 second, allowing rapid transaction execution with state preconditioning and changing.
```
teddyjfpender, line 9: This is related to not just smart contracts but more generally block-time, correct?

RFCs/rfc-0004-fast-transactions.md
```diff
@@ -0,0 +1,77 @@
...
+
+- Non-use of preconditions and nonce manipulation (for example, in the case of the stateless contract).
+- Recursive proofs via ZkProgram for bundled transactions.
+- Using Actions and Reducer
+
+These methods, especially recursive proofs, pose significant entry barriers due to complexity and infrastructure demands and do not allow for many uses, as in the case of Actions and Reducer.
+
+**Problem**: The Mina blockchain's slow transaction pace (3 to 20 minutes) hinders rapid processing, contrasting with other blockchains like Tron (3 seconds) and Ethereum (12 seconds).
+
+**Relevance**: Accelerating transaction speed can greatly enhance the blockchain's Total Value Locked (TVL) and user experience, fostering Mina's ecosystem growth.
```
teddyjfpender, line 31: Can you define who the core consumer of blockspace would be as the block-time becomes faster? I personally believe that the consumer of blockspace who uses actions + reducers or transactions on the L1 is likely very different from the consumer of blockspace who uses off-chain recursive proofs and ZkPrograms and submits a single state update. 

It would be great to outline why the idea for faster block-time is applicable, useful, and important to each different consumer persona.

RFCs/rfc-0004-fast-transactions.md
dfstio: You are correct, but other use cases are already possible without this RFC. You can now calculate recursive proofs with ZkProgram and include many proofs in one block. You can send many non-contract transactions by increasing the nonce by one each time, and all those transactions will be included in the same block.

What is different with SmartContract is that to include the transaction to the block, the precondition is checked. This precondition now is checked against the last block account state, so you cannot include several transactions that rely on different precondition states, you should wait for the next block.

RFCs/rfc-0004-fast-transactions.md
dfstio: I would consider two types of customers to give the correct answer to your question: end customers and developers.

**The end customers** hardly know and do not care what is under the hood, but they do care about two things:
- Speed of sending the transaction
- Cost of sending the transactions
You can see how Tron TRC-20 tokens took the leading role in money transactions mainly because the transfers are fast and cheap, and Tron TVL is now 4 times higher than Polygon TVL, notwithstanding the fact of $2B investments into Polygon and many advanced ZK technologies integrated into Polygon.

It applies to all types of applications. The attention span of the people now is 8 seconds, and being able to deliver the result within those 8 seconds leads to a conversion increase. If the result is delivered after 8 seconds, many people have already forgotten what they would be doing and switched their attention to the next thing.

**The developers** are looking first to satisfy the end customers, so the above is of utmost importance for them. 

They have a wide range of tools available to them, and one of the tools they have used in recent years, as I see from my experience, is not to use the blockchain at all and base the design on old architecture to meet their business objectives, so the first factor we are speaking about is the adoption of the blockchain technology. 

The other tools as L2, recursive proofs, Actions/Reducer are available for them, and they will choose the tool on the
- Easy to use
- Ability to meet the objectives
- Cost
- SLA levels

In this sense, all those solutions are complementary. Actions/Reducer is easy to use and does not require external infrastructure and additional cost, but you cannot update Merkle Map with it.

Recursive proofs are very powerful but require a significant CPU and memory. Fast calculation and merging of even 32 proofs in a reasonable time requires the computer to have the parameters that exist only in a cloud/server environment. Setting up such an environment is difficult for the developer who just came to the ecosystem.

L2 is probably the best solution, but the questions of data availability, messaging, and asset transfer between L1-L2 are always complicating the development.

The experienced developer can handle all this, but if we are speaking about a junior developer, he is looking to implement his idea based on a slight modification of the example from the tutorial and is not looking to integrate with L2 or cloud infrastructure except when this integration is very simple and already included into example))

The focus of this proposal is to provide capabilities equal to L2 but without the need to change the tutorial code or do some integrations. In some sense, you can consider this proposal as a “light” L2 with a life span of 3 minutes that, at the end of its life, as a new block comes, transfers all the assets and state to the L1 and disappears.

To summarize, this proposal is essential for applications that have to make transactions fast and cheap without significant CPU and memory requirements.

The example of such applications are:
- Games
- Trading (and to withdraw safety, you still need to wait for a block, but trading can be done within the block)
- Putting live data into the blockchain in real-time (oracles for quotes, legally important audio, other real-time data streams)
- Real-time voting in the meetings
- Any transactions where identity should be present to the zkApp and relied on to make the transaction, such as including the person in a white list before making the transaction.

dfstio: speed

teddyjfpender: You raise many fantastic points here. Please include this in the RFC's markdown!

In my opinion, the primary challenge to scaling Web3 applications for many users is eliminating the system complexity that manifests in user experiences. Congestion, high fees, and long waiting times are just a few of the complexities of a blockchain that many users experience in a Web3 application. One way to mask these complexities is with higher TPS or similar workarounds, but these seldom solve the problems, only delaying them. Still, these complexities do not exist in end-users mental models of the Web3 application they're interfacing with, and they arise in their interaction with Web2 applications. Ephemeral zkVMs are one interesting way to address these challenges with the first-in-class features Mina has.

I'm keen to see if others have thoughts in this direction too.

mrmr1993 (changes requested):
I don't believe this is technically realistic / feasible as written. See comments for concerns and requested changes.

RFCs/rfc-0004-fast-transactions.md
maht0rz: The zkVM that @teddyjfpender described above is already implemented by Protokit, you can start using it today at https://protokit.dev

Realistically we can do between 3-12 TPS with some minor changes to the underlying protocol, we can also do *instant* transaction inclusion confirmation (via receipts), data availability via sequence state or events and many other things that can be added to the core protocol.

Features available currently are:
- Abstractions over writing zk programs that manage state via state transitions, 
- Soft-failing transactions, inclusion of failed transaction in a provable way into a block
- Key-value storage a.k.a. "infinite" state
- Out-of-prover sequencing of transactions
- Parallelisation of proof generation in block production (by separating tree operations and business logic)
- Development environment with support for UI - via GQL endpoints and a handy CLI

Features that are currently being implemented:
- L1/L2 bridge
- Sequence state mempool, as opposed to the private mempool used currently
- Indexing & processing of historical data
- and many more features coming soon! :)

I'd encourage anyone facing the issues described in the RFC to try out our starter kit and join our discord, here's a list of useful links and resources:
- https://github.com/proto-kit/starter-kit
- https://protokit.dev
- https://discord.com/invite/AMGnGAxsKp
- https://www.youtube.com/watch?v=hqobPb3gIRo

RFCs/rfc-0004-fast-transactions.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Scenarios and Use Cases
+
+### Scenario 1: High-Frequency Transaction SmartContract
+
+Example: A chess blitz game SmartContract needing several transactions per minute. This feature would enable high-frequency transactions without complex proof infrastructure.
+
+### Scenario 2: Multi-User SmartContract Requiring Immediate Final State
+
+Example: A multi-user SmartContract that operates on simultaneous requests and requires immediate access to the final state post-transaction. It can use Actions and Reducer to collect user requests, then send a reduce transaction and receive the final state to be used in the following user's transactions immediately.
```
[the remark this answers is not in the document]
dfstio: I've amended the RFC, taking into account your comments. The new version can be seen on
https://github.com/MinaFoundation/Core-Grants/blob/89b16a6c6d7b307a55ccad3d281eb96e0c35a3cc/RFCs/rfc-0004-fast-transactions.md

Please check it and let me know of any additional comments, questions, critiques, or concerns. Your feedback is highly appreciated.

RFCs/rfc-0004-fast-transactions.md
dfstio: I've included those details in the new version of RFC published today.
I've also included a detailed description of the guaranteed state using the contract terms to make it clear and have introduced a new term, "calculated state".

RFCs/rfc-0004-fast-transactions.md
dfstio: I've included the chart above in the new version of RFC. Thank you, Teddy, for it!
I've also added several new diagrams to illustrate RFC concepts.

The information about possible TPS above is very interesting, I've included the section "Lower theoretical bound of the interval between transactions and why it matters" to the RFC to discuss the lower possible bound of the interval.

RFCs/rfc-0004-fast-transactions.md
dfstio: I've included more detailed opinion about the Action/Reducer. It is a great tool and I've also included into RFC some ideas on how it can be enhanced with a disclaimer that it is not a scope of the current RFC. There is also good discussion on discord on Actions and Reducer:
https://discord.com/channels/484437221055922177/1173407599123767346

I've amended the title and introduction in accordance with your recommendation. Thank you for it, I really like your wording!

RFCs/rfc-0004-fast-transactions.md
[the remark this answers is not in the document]
dfstio: I've also created an example RealTimeVoting zkApp to show how many transactions zkApp can prepare and send in a short time. You can see the 122 RealTimeVoting  zkApp transactions included in TestWorld2 block 18241:
https://minascan.io/testworld/block/3NKCH5552TNFAGhCbprddsbSbLwc4YqimqLVjeepY9Hg2Kh9Darf/zk-txs

This zkApp does not use recursion over proofs, but it is able to prepare proof every second using integration with the MinaNFT backend:
https://github.com/dfstio/minanft-api/blob/rfc4/src/api/rfc4.ts

 I've also introduced the term "calculated state," which allows for solving the problem of proof recursion by splitting state calculation process that cannot be parallelized and proof/recursive proof calculation process that can be run in parallel with almost every proof per second rate achievable due to parallel processing.

For example, you can see my statistics on recursive proofs that I've published on discord:
https://discord.com/channels/484437221055922177/915745847692636181/1182624854973874186
There, I calculate and merge 32 proofs in 7 minutes using parallel processing. It is 13 seconds per proof. This time can be made shorter if the number of proofs is higher.

Those two methods, using calculated state and parallel proof processing, have the potential of 1-second transactions. 

I've also included the section "Lower theoretical bound of the interval between transactions and why it matters" to discuss the lower theoretical bound of the interval.

RFCs/rfc-0004-fast-transactions.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Introduction
+
+Mina Protocol's SmartContract transactions, authenticated through proof, must meet these criteria:
+
+- Matching account update preconditions with the current account state.
+- Proof verification against the SmartContract's verification key.
+- Fee alignment with the market rate.
+
+Once verified, the transaction's inclusion in the block is assured, bypassing the need for blockchain-wide recalculations. Despite a known post-verification state, SmartContracts must wait about 3 minutes for state reflection in the new block before initiating another transaction.
```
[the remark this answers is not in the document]
dfstio: Agree; I've amended the RFC.

If we are speaking about one-actor transactions (that also include the case when zkApp is the sequencer for user requests), we should also note two limitations:
- The total limit of zkApp transactions in one block is around 120.
- The actor cannot simply increase the nonce and send the new transaction to the same block in case the transaction includes preconditions checks and changes the state.

This RFC aims to shift those two limitations.

RFCs/rfc-0004-fast-transactions.md
```diff
@@ -0,0 +1,77 @@
...
+
+- Current system limitations: Transactions can take 3 to 20 minutes.
+- Real-world applications: Many scenarios require transaction intervals of only a few seconds.
+- Ecosystem growth: Faster transactions open up more use cases and attract developers.
+
+## Possible Implementations
+
+### Variant 1: Node-side Recursive Proof Computation
+
+Node, after transaction verification, computes the new account state, sign it, and returns it to the zkApp. The zkApp can utilize this state (for example, by providing it as an argument to the fetchAccount()) to create new immediate transactions. The node can replace the existing mempool transaction with a new composite one, reflecting the initial transaction's preconditions and the final state of the last transaction.
```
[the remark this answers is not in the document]
dfstio: Agree. I've removed the concept of the node signing the guarantee state from the RFC and made explicit that this RFC is speaking about centralized ephermal L2 blockspace.

I've also added explicit wording about sequencer roles in different scenarios and added the section on creating compound transactions. 

This section does not answer all the questions, so I've added this issue on creating compound transactions into Open Issues and Discussion Points section.

RFCs/rfc-0004-fast-transactions.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Variant 2: zkApp-side Recursive Proof Computation
+
+SmartContracts gain the ability to merge multiple transaction proofs. The first transaction proceeds normally, with subsequent transactions being based on the internal state maintained by o1js and the calculation of the merged proofs. The composite transaction, combining initial preconditions, merged proof, and final state, is sent for replacement to the node.
```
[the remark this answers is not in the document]
dfstio: I've removed the o1js recursive proof calculation section and instead added a discussion about "calculated state" and compound transaction creation.

RFCs/rfc-0004-fast-transactions.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Scenarios and Use Cases
+
+### Scenario 1: High-Frequency Transaction SmartContract
+
+Example: A chess blitz game SmartContract needing several transactions per minute. This feature would enable high-frequency transactions without complex proof infrastructure.
```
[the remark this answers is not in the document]
dfstio: Added

RFCs/rfc-0004-fast-transactions.md
```diff
@@ -0,0 +1,415 @@
...
+
+Our goal in Web3 is not to maximize the number of blockchains. Maximizing the number of blockchains is something I would state explicitly as a non-goal, as it primarily benefits validator cartels seeking to extract value. Our goal in Web3 is to maximize the amount of blockspace that exists and ensure it is allocated to the state machines which need it most at any time: a constant generation and allocation of global consensus resources to those who need it the most. An enterprise without waste. In other words: the most effective blockspace producer in the world.”
+
+### BlockSpace Node
+
+BlockSpace Node is a node that, in addition to the usual services of the Mina protocol node, provides the services for running blockspace - an L2 epithermal blockchain. It is up to the node to decide whether it wants to run the blockspace service, whether it is being provided to every zkApp or just to the zkApps whitelisted by this specific node, and what the fees are for running blockspace services.
```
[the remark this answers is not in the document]
maht0rz: I'd tend to agree with what @teddyjfpender has said above, what you describe as a "BlockSpace node" is just a variant of how an L2 can be implemented on top of the MINA L1. Furthermore the RFC is missing technical details regarding the mempool usage or "guarantees".

Regarding the guarantees for client-side zkApps, you can have these features today using Zeko - at least from what i understand from their whitepaper & discussions with the team. Of course given you want to stick to the AccountUpdate model. The incentive for the Zeko sequencer operator is pretty straightforward - they get to keep the difference between what they generate in transaction fees and the cost of settlement to the L1. 

Both Zeko and Protokit allow developers to run their existing sequencing software to address the issues you've raised in this RFC. What is missing in Zeko or Protokit in your view, that is mitigated by the design described in this RFC? I personally don't see any difference in what you described vs how Zeko & Protokit operate today, and thats the main reason why i ask ^^.

Another thing i'm not entirely sure about is why you chose to skip actions / sequence state from the RFC entirely, in my understanding thats the most powerful primitive (besides cheap settlement) the L1 offers for L2s. It allows anyone to tap into the L1 consensus and use the L1 as a mempool - commonly referred to as based sequencing. Based sequencing allows the L2s to become decentralized - unlike a centralized "Blockspace node".

To address the concerns about running your own infrastructure for an L2 as an entry barrier - you don't have to do that with Zeko, and you won't have to do that after (if) we launch a general purpose rollup with Protokit either. You have to run your own infrastructure in case you want to run your own instance of Zeko, or your own instance of Protokit for an application-specific chain. I'd argue this is actually a desirable property for further decentralisation and scaling, as we know general purpose rollups or centralized infrastructure providers (like Infura) are sub optimal for the network health.

There are some additional concerns i have e.g. as you've described the depositing of assets back and forth between the short lived L2 - you're going to be limited by the amount of AUs you can pass in a single transaction today. That means if you end up merging multiple AUs together to chain their root hash state from A --> B, you won't be able to keep chaining all the other functionality that the AU offers, unless of course you design the L2 like Zeko did. Which means re-using the L1 zkapp command circuits to maintain 1:1 functionality when recursively merging AUs into your L2 state.

Apart from all the things i've touched upon above, there's a dozen different things you have to consider when designing a "VM" for a rollup or an L2. I know i'm repeating myself - but we've covered the architecture behind Protokit in our [recent workshop](https://www.youtube.com/watch?v=hqobPb3gIRo). Proof merging, transaction processing parallelisation, state transition proving parallelisation, user-land abstractions over state transitions, those are all features available today in the existing Protokit SDK. Is there anything specific thats missing from the Protokit stack to complement this RFC? Given the modular nature of the SDK, we'd be more than happy to accommodate for any missing features you identify.

RFCs/rfc-0004-fast-transactions.md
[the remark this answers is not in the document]
mrmr1993: > There, I calculate and merge 32 proofs in 7 minutes using parallel processing. It is 13 seconds per proof. This time can be made shorter if the number of proofs is higher.

This seems at odds with the new wording in the abstract:

> describing a light ephermal L2 blockspace with a life span of about 3 minutes that

In particular, all 32 transactions could have been processed faster by just including them in the L1 block. You need over 380 txns for the throughput to be higher than Mina, which implies a parallel proving time of about 12 mins, and presumably a corresponding life span for the 'ephemeral' blockspace.

RFCs/rfc-0004-fast-transactions.md
```diff
@@ -0,0 +1,415 @@
...
+
+  - The transaction is included in the L1 block
+  - ZkApp calls off the transaction by sending the replacement transaction
+  - The transaction is excluded from the mempool due to the low fee or due to the maximum time reached for the transaction to be in the mempool as the result of the violations by zkApp of the Condition Precedent.
+
+- **Rights of Third Parties**: Third Parties, including, but not limited to, other zkApps and Accounts, are NOT beneficiaries of the State Guarantee Contract and can rely on this guaranteed state only in the optimistic sense by also relying on the architecture of the zkApp and not on this Guarantee. It is recommended for third parties to wait for BlockSpace to commit to the L1 before executing any material transaction that relies on the state, such as asset withdrawal.
+- **Applicable law**: Mina protocol
+- **Arbitrage**: O(1)Labs
+- **SLA**: TBD
+
```
mrmr1993, line 169: This is very odd, and still doesn't answer any of the essential questions of the guarantee that make it useful. It would be better to delete this and add answers to the following:
* What does the `BlockSpace` node do?
* How do users or the zkApp choose/know which node to go to?
* What are the censorship consequences of this design?
* How does a block producer find out about the `BlockSpace` node's sequenced transactions?
  - Please include specific technical communication mechanisms.
* Why would a block producer include the `BlockSpace` node's sequenced transactions?
  - Technical communication mechanisms needed here too: how does it know that there is a `BlockSpace` node? How does it know there's not more than one racing? How does it know that there are pending transactions, and so it shouldn't execute a one-shot txn for the zkApp? etc.
* Why wouldn't a block producer include some non-`BlockSpace` node's transaction for a zkApp and invalidate all of the work?
  - Specifically, please talk about the incentives. The Mina block producer node would rather maximize their fee; if the `BlockSpace` node's bundle isn't ready when their slot comes around, they can sacrifice the fee to preserve the bundle, but most will just take the fee and ruin all the `BlockSpace` node's work.

RFCs/rfc-0004-fast-transactions.md
```diff
@@ -0,0 +1,415 @@
...
+- **Term**: The time that starts when zkApp sends the transaction to the BlockSpace Node to be committed to the L1 blockchain and ends on the earliest of:
+
+  - The transaction is included in the L1 block
+  - ZkApp calls off the transaction by sending the replacement transaction
+  - The transaction is excluded from the mempool due to the low fee or due to the maximum time reached for the transaction to be in the mempool as the result of the violations by zkApp of the Condition Precedent.
+
+- **Rights of Third Parties**: Third Parties, including, but not limited to, other zkApps and Accounts, are NOT beneficiaries of the State Guarantee Contract and can rely on this guaranteed state only in the optimistic sense by also relying on the architecture of the zkApp and not on this Guarantee. It is recommended for third parties to wait for BlockSpace to commit to the L1 before executing any material transaction that relies on the state, such as asset withdrawal.
+- **Applicable law**: Mina protocol
+- **Arbitrage**: O(1)Labs
```
mrmr1993, line 167: I don't think you should add this here without seeking approval from O(1) Labs first. O(1) are not formally involved in this process (to the best of my knowledge).

RFCs/rfc-0004-fast-transactions.md
```diff
@@ -0,0 +1,415 @@
...
+
+### The calculated state
+
+zkApp can calculate the state that will be the result of the transaction in a way similar to how Mina.transaction() does (or by using Mina.transaction() without calling transaction.prove())
+
+This calculated state can be used immediately by the zkApp to start creating and proving the next transaction without waiting for the previous transaction.prove() to be finished, enabling parallel proof calculations.
+
+The calculated state is not guaranteed and can be considered simply as one of the variables zkApp uses. To use it, Mina.transaction() arguments have to be amended to accept calculated or guaranteed state to be used as precondition input instead of the default precondition account state returned by fetchAccount() and referring to the last block account state.
+
+The calculated state will become a guaranteed state in an optimistic sense and, later, a final state in an optimistic sense.
```
mrmr1993, line 178: This, and the previous 2 sections, could be replaced with the sentence `We will optimistically assume that, whatever the BlockSpace node claims that the zkApp's state will be, is the new state that we should build upon.`

You can then use the space you've saved to describe how this will handle:
* any state changes while a zkApp proof is in flight, e.g.
  - txn1 requests the state, receives S1, begins proving
  - txn2 requests the state, receives S1, begins proving
  - txn1 finishes proving, submits, state becomes S2
  - txn2 finishes proving, its proof is now invalid.
  - (Scale this to the potential number of concurrent executions in the system.)
* cross-sequencing with other zkApps
  - Concretely, a zkApp may 'call' another zkApp, and thus depend on the linked state of the 2 zkApps.
  - This may become arbitrarily complex, if a zkApp is called by or calls may different other zkApps.
  - If 2 zkApps attempt to interact, but they are being sequenced by 2 different `BlockSpace` nodes, which node do they go to? How do those nodes coordinate? Do they coordinate? How does the block producer who might include one or other of the bundles know which to inlcude?
* communication
  - What do the queries look like from o1js to the `BlockSpace` node?
  - What is the incentive to serve those queries?
  - How does the `BlockSpace` node communicate with the upcoming block producer? (NB: that they are unknown until the block has been produced)
  - Concretely, running through multiple scenarios of racing transactions etc., run through which messages are sent to and from which entites at which time, what calculations they run, and what failure cases can occur at each of these interactions.

mrmr1993 (changes requested):
This is still lacking on technical details. I've added a few more questions/comments/suggestions, but broadly I agree with @maht0rz's [comment](https://github.com/MinaFoundation/Core-Grants/pull/7#discussion_r1427014171), and especially the question:

> Both Zeko and Protokit allow developers to run their existing sequencing software to address the issues you've raised in this RFC. What is missing in Zeko or Protokit in your view, that is mitigated by the design described in this RFC?

I don't think the proposal here is technically sound or feasible as described. My preference is to close this RFC without funding, given those technical issues, and that there is no clear path to resolving them.

RFCs/rfc-0004-fast-transactions.md
dfstio: @mrmr1993 
Thank you for your prompt answer and comments, which are always very valuable and to the point.

I have very limited documentation available for both Zeko and Protokit. Zeko is not available yet for testing, and Protokit is available with proofs off, so I prefer not to compare their features at this stage due to the lack of information.

This RFC has raised several issues for community discussion, and not all of those issues are technical. I believe that we should first look at the business objectives, and the technological decisions should be aligned with them.
- The thesis regarding the importance of fast, 1-second transactions for Mina protocol adoption. Is this of utmost importance for Mina protocol adoption, or merely a nice-to-have feature?
- If fast transactions are critical and required for Mina protocol adoption, and they are not built now in the protocol, can L2 solutions like Zeko and Protokit fully address this need? Or do we need to add it to the protocol or have a broader range of solutions aiming to have at least three solutions for fast transactions?
- Zeko is a full-fledged L2 with bridge, DA layer, explorer, etc. Do we need a “light” ephemeral blockspace that will complement this L2 solution or not?

Depending on the consensus on answers to the above questions, we must continue to find technical design decisions for light ephemeral blockspace or other solutions for fast transactions and integrate it into the Mina protocol or make it a separate solution, with this discussion on the RFC being a good start, or we may close this topic if it's deemed non-essential for Mina protocol adoption.

I would highly appreciate your opinion on the above questions as you undoubtedly have more information and insights on those topics, and your opinion is very valuable for the community.

maht0rz: @dfstio Protokit isn't only available with 'proofs off', we simply did not put any emphasis on the feature yet since we haven't tested the entire framework with proofs enabled end to end. But to address this i wrote a test that runs a sample Protokit runtime `Balances` in the prover. It takes roughly 20s to prove a circuit of < 500 gates, this circuit is an implementation for `transfer()` which internally handles 4 state transitions, uses `this.transaction.sender` and `assert()` APIs of Protokit (end to end provable). Furthermore the state transitions need to be proven in parallel, and will be proven in batches of 4 in state transition circuits that almost exhaust the 2^16 constraint limit. Proving a 2^16 circuit in wasm takes about 50s on my machine. 

The way Protokit is architectured allows running both provers in paralel, therefore it takes about 50s combined to prove  + then to merge the proofs of runtime execution and state transition application (2x verify() in our BlockProver).

You don't have to write any other code than your runtime business logic, rest of what i've described including sequencing, mempool, proof paralelization / task input precomputation etc is handled internally by the framework.

Here's the code for the Balances runtime:

```typescript

@runtimeModule()
class Balances extends RuntimeModule<Record<string, never>> {
  @state() public balances = StateMap.from<PublicKey, UInt64>(
    PublicKey,
    UInt64
  );

  @runtimeMethod()
  public transfer(from: PublicKey, to: PublicKey, amount: UInt64) {
    // read current balances
    const fromBalance = this.balances.get(from).value;
    const toBalance = this.balances.get(to).value;

    const fromIsSender = this.transaction.sender.equals(from);
    // check if from is the sender
    assert(fromIsSender, "Sender mismatch");

    // check if balance is sufficient
    const fromBalanceIsSufficient = fromBalance.greaterThanOrEqual(amount);
    assert(fromBalanceIsSufficient, "Insufficient balance");

    // naive implementation of 'safe sub' thats circuit friendly
    const paddedFrombalance = fromBalance.add(amount);
    const safeFromBalance = Provable.if<UInt64>(
      fromBalanceIsSufficient,
      UInt64,
      fromBalance,
      paddedFrombalance
    );

    // calculate new balances
    const newFromBalance = safeFromBalance.sub(amount);
    const newToBalance = toBalance.add(amount);

    this.balances.set(from, newFromBalance);
    this.balances.set(to, newToBalance);
  }
}
```

You can find the complete test for the runtime in-prover in our repository:
https://github.com/proto-kit/framework/blob/feature/runtime-zk-programmable/packages/module/test/RuntimeZkProgrammable.test.ts#L220

P.S.: We're actively working on documentation, in the meantime the entire repository is open source and can be learned from e.g. by checking out the provided tests and examples. I think its not exactly productive for everybody involved to write a whole RFC describing a solution that's by design worse off than the existing ones, with the main argument being "docs dont exists" (for both). If there's anything missing i'd be more than happy to explain in any format desirable (office hours, calls, workshops) until the full docs are ready.

TyrellCorp2020:
Not much activity lately, let's move to draft as it seems infeasible at present.

PR #9, opened 2023-12-26
RFC-0005: Protocol Upgrade to Enable Automatic Distribution of Delegation Rewards

lampardlamps:


teddyjfpender:
Thanks for making this PR @lampardlamps! 🙏

RFCs/RFC-0005: Protocol Upgrade to Enable Automatic Distribution of Delegation Rewards.md
```diff
@@ -0,0 +1,49 @@
+# RFC-0005: [Protocol Upgrade to Enable Automatic Distribution of Delegation Rewards]
+
+- **Intent**: [To request a protocol upgrade for automatic distribution of delegation rewards.]
+- **Submitted by**: [lamps, discord: lamps6718, twitter:lamps945 ]
+- **Submitted on**: [26 Dec 2023]
+
+## Abstract
+
+This RFC aims to address one big pain point affecting all node operators on Mina network, which is the distribution of delegation rewards. The current mechanism relies the node operators to voluntarily calculate the rewards and send them back to the delegators. It is error-prone for the calculations and, more importantly, relies complete trust of the node operator, which hinders the decentralisation of the network. The RFC requests a redesign of the mechanism and achieve one that is similar to the Cosmos chains of Cardano, where the rewards are not accessible to the node operators, and the delegators can check and claim the rewards any time.
```
teddyjfpender, line 9: > The RFC requests a redesign of the mechanism

Core RFCs are not for requesting artefacts like a "redesign" directly but rather to conjure up artefacts that allow one to produce such an artefact by outlining scenarios, and requirements as well as engaging in discussion to refine those items. Please reword this last sentence 🙏

RFCs/RFC-0005: Protocol Upgrade to Enable Automatic Distribution of Delegation Rewards.md
```diff
@@ -0,0 +1,49 @@
...
+
+Two major issues come with this relatively rudimentary design: one is the congestions of the network at the end of each epoch when the snapshot of ledger is taken, which resulting in most operators choosing to send transaction during such times. The other is that the Mina holders will only delegate to the operators who they can completely trust, to avoid risking loss of rewards, which hurts the decentralisation of the protocol and results in the funds being concentrated to a small number of reputable operators, instead of having a long tail distribution of decent-sized operators and enthusiasts.
+
+## Objectives
+
+A redesign of the mechanism and achieve one that is similar to the Cosmos chains of Cardano, where the rewards are not accessible to the node operators, and the delegators can check and claim the rewards any time.
```
teddyjfpender, line 19: > similar to the Cosmos chains of Cardano

Do you mean "Cosmos chains _or_ Cardao"?

RFCs/RFC-0005: Protocol Upgrade to Enable Automatic Distribution of Delegation Rewards.md
```diff
@@ -0,0 +1,49 @@
...
+
+Two major issues come with this relatively rudimentary design: one is the congestions of the network at the end of each epoch when the snapshot of ledger is taken, which resulting in most operators choosing to send transaction during such times. The other is that the Mina holders will only delegate to the operators who they can completely trust, to avoid risking loss of rewards, which hurts the decentralisation of the protocol and results in the funds being concentrated to a small number of reputable operators, instead of having a long tail distribution of decent-sized operators and enthusiasts.
+
+## Objectives
+
+A redesign of the mechanism and achieve one that is similar to the Cosmos chains of Cardano, where the rewards are not accessible to the node operators, and the delegators can check and claim the rewards any time. 
+
+This will help resolve both issues mentioned above: the reward claim transactions can be much evenly distributed in time during each epoch (instead of congestion towards the end/beginning), and the necessity to trust a node operator is removed.
```
teddyjfpender, line 21: Another objective could be to remove as much system complexity as possible from the end-user's experience, in this case, a user staking MINA in a wallet. In my opinion, that user should not be concerned with a "claim" transaction at all; rewards should always be spendable and can be used for things like paying transaction fees for example (reducing two transactions into one -- one claim, one spend -> one spend).

A scenario where a wallet user would need to be aware of a "claim" transaction is if there were constraints around the distribution of rewards, notably, this was a large topic in 2023 in the Cardano ecosystem; [contingent staking](https://forum.cardano.org/t/personal-thoughts-on-contingent-staking/114912). I wonder if there is a case where the Mina ecosystem believes a requirement for this feature would be to enable constraints on the spending (i.e. claiming) of staking rewards, in essence, to give those accounts programmable logic.

es92:
Just wanted to surface here, I think there was a previous plan, to (1) add binding coinbase receiver to zkApps, and then (2) have a zkApp that can manage payouts (probably yeah based on the work Gareth had done). Do we think that would be sufficient, or is more necessary? Building a complete version of that zkApp could probably be a good grant to come out of this (and something that could go into production after hard fork launch, though a fully constrained version would have to wait until coinbase receiver is added in)

lampardlamps:
> Just wanted to surface here, I think there was a previous plan, to (1) add binding coinbase receiver to zkApps, and then (2) have a zkApp that can manage payouts (probably yeah based on the work Gareth had done). Do we think that would be sufficient, or is more necessary? Building a complete version of that zkApp could probably be a good grant to come out of this (and something that could go into production after hard fork launch, though a fully constrained version would have to wait until coinbase receiver is added in)

Hi Evan, thanks for your comments. I was not involved in the details of the discussion. Was it determined that implementing a zkApp for this purpose would be the better choice than one on the protocol level? If so, the zkApp should be okay as long as it satisfies the requirements outlined here, namely: 1. the block producer do not have direct access to the block rewards; 2. the delegation charges are configurable onchain; 3. the rewards for each user is directly distributed/ available for claim according to the charge configurations.

lampardlamps:
@teddyjfpender thanks for your comments Teddy, I've made some changes accordingly, I hope they are acceptable.

es92:
> > Just wanted to surface here, I think there was a previous plan, to (1) add binding coinbase receiver to zkApps, and then (2) have a zkApp that can manage payouts (probably yeah based on the work Gareth had done). Do we think that would be sufficient, or is more necessary? Building a complete version of that zkApp could probably be a good grant to come out of this (and something that could go into production after hard fork launch, though a fully constrained version would have to wait until coinbase receiver is added in)
> 
> Hi Evan, thanks for your comments. I was not involved in the details of the discussion. Was it determined that implementing a zkApp for this purpose would be the better choice than one on the protocol level? If so, the zkApp should be okay as long as it satisfies the requirements outlined here, namely: 1. the block producer do not have direct access to the block rewards; 2. the delegation charges are configurable onchain; 3. the rewards for each user is directly distributed/ available for claim according to the charge configurations.

Yep I think it would match those requirements - and yep the thought is if it can meet requirements it would be better this way than at the protocol level, since this way features can be added / removed as needed by BPs / staking pools, and without requiring HFs.

lampardlamps:
> Yep I think it would match those requirements - and yep the thought is if it can meet requirements it would be better this way than at the protocol level, since this way features can be added / removed as needed by BPs / staking pools, and without requiring HFs.

Thanks Evan, makes perfect sense!

EmrePiconbello:
I think Gareth should get involved in this conversation. I was a follower of that project and the last update I got "because of reliance on oracles. It's not worth it because of risks" I might be recalling details wrong.

garethtdavies:
> I think Gareth should get involved in this conversation. I was a follower of that project and the last update I got "because of reliance on oracles. It's not worth it because of risks" I might be recalling details wrong.

The big blocker was the lack of coinbase receiver binding - as anyone could just switch the coinbase receiver and not use the zkApp, which makes the solution kind of moot. As I understand at, this time, there won't be an option to make the contract immutable i.e. `permissions: None` which again means that it wouldn't be permissionless (I think this is just temporary as the crypto is likely to change). Finally, as this relies on an oracle, this would need to be very robust - I'd started some research on recursively validating multiple signatures from different oracles (circuit size was an issue otherwise, though things may have changed). So my comments there were more around launching something like this right after the hard fork, which isn't currently very feasible, rather than long-term.

jrwashburn:
While the theory of payout creativity was an exciting idea, the reality is that it's been pretty limited. This is not some application-layer capability - it's directly related to the security of the blockchain and should be treated as such. I don't think there is much upside to a bunch of creativity in this area. A zkapp/oracle-dependent solution is a workaround, and while better than our current work-arounds, this should be implemented by the protocol.

EmrePiconbello:
@michal0mina 
I am fully supportive of an on-chain protocol-level solution. I understand it's the most complicated option, and given our limited block space, it's also quite challenging. However, here's the thing: to my knowledge, since 2017, no PoS blockchain relies on validators for reward distribution. Anyone can claim or distribute rewards, or it's fully automated. Additionally, all rewards go directly to their intended recipients without the need for trust. As regulations continue to develop, and PoS operations become more clearly defined, Mina doesn't fit the standard PoS model in many cases. Custody is also a concern, with many regions reviewing it due to FTX. We're not part of the foundation's delegation program because of these issues. While this isn't a major concern, it's difficult to justify when all PoS chains function similarly, while Mina operates differently. A significant portion of the problem stems from custody, with many regions requiring e-money processing licenses (commonly used by fintech firms and challenger banks). This could lead to many validators being forced to exit Mina in the long run or face a 20% reduction in rewards due to taxes, or potentially face various charges since we are directly involved in signing transactions, making us liable under the law. That's why we need a trustless system where custody, distribution, calculation, etc., don't rely on validators. 

From what I outlined above I could only support the first option while I didn't get the daemon part of that solution. I assume because no node actually hold data we are configurating the daemon to hold data for last 2 epochs but if you can elaborate more that would be very helpful.

michal0mina:
@EmrePiconbello Thanks for the comment and legal insights. I incorporated them in the big comment


Re: the daemon solution -- this is it, but I mentioned those low level details as it's serious sub-system and long term maintenance, both for developers and node operators (as it is heavy system).

jrwashburn:
Regarding the cons of a daemon solution: The daemon already has the entire delegation structure for the current and next epoch. The only thing to add (data-wise) would be the block producer fee. 

Separately, having the rewards be claimable, vs. auto-incrementing, would be very useful from a tax perspective to give delegators the ability to control timing of receipt of rewards. While this isn't anticipated in what @michal0mina has written above, it is something to consider. This feature would add significantly more complication due to a) wanting to consider the rewards in pos lottery, and b) providing an accumulator to keep unpaid rewards. I don't have a strong opinion re: adding this to the specification, but it should be considered given its prevalence in other protocols.

michal0mina:
@jrwashburn Thanks

Re the daemon -- I will investigate what we have now. Yes, we have the delegation ledger, but I thought it is only on BP level.
Let me double check it.

About the logic that you describe:
* it's basically choice between distributing delegation and accumulating delegation 
* I worked on a project that allowed delegators to specify when delegating which policy they prefer 
* I will add it as nice to have to the protocol solution

jrwashburn:
> Let me double check it.

The only place to get the staking ledgers (at least that I am aware of) is to export them from the daemon using `mina ledger export [staking-epoch-ledger | next-epoch-ledger]` (unless they are archived by someone who has done that - a la https://minaexplorer.com/staking-ledgers

jrwashburn:
For the daemon solution, I think we need to consider:

New transactions:  

1. Register Block Producer Pool - need to be able to specify Pool Fee

2. Unregister Block Producer - consider whether this is necessary, and if so, should it change the delegation of all delegates back to themselves? This transaction would allow all delegates to be aware that the pool is exiting block production (vs. simply not producing blocks.) 

We could also consider an Update Fee transaction; however, I think it is best to unregister and re-register with the new fee, so that delegates have a clear choice to accept the new fee amount. A further complication to consider would be supporting update fee - but it only taking effect for delegates after a certain date. This all seems like more trouble than it is worth and I suggest the simpler register/unregister approach. I'm certain the community will come up with monitoring solutions to alert delegates of the unregistrations, and expect wallets would notify on unregister as well.

Another advantage of this approach would be that it could create an official registry of pools / those BP's offering to accept delegations. (Today, you can delegate to anyone, and we have seen instances of people setting up a pool fee of 100% to make it more clear that the are not a pool. This registry could be used to prevent delegating to people that do not want delegations.)

michal0mina:
@jrwashburn re the delayed update fee, it would be reasonable to apply at least the same delay as for delegations (epoch+1). If there is need for signalling, it may be increased. Also, it's worth considering adding constrains (e.g. max 2x increase, or max 50% increase)

Lack of update mechanism would be painful for managing ecosystem of your users. Some sort of inheritance of the old delegators is surly helpful.

Sventimir:
I'm no expert in legal issues, but from purely technical point of view the simplest solution is to bake it into the protocol. Very simply, whenever the node constructing a block pays themselves a coinbase reward, it also scans the staking ledger in search of accounts delegating to them and immediately distributes the rewards among them. It may or may not add explicit transaction for each thus paid reward. While this is not necessary from the protocol's perspective (validators can compute those rewareds for themselves from the information already included in the block), it'll make it harder for external software like indexers or Rosetta to figure out what's going on in the ledger (they'd basically need to replicate the logic).

Putting this into a zkApp, while probably could be done without changing the protocol, will require someone to run the zkApp, so it's still up to a BP whether or not they run the zkApp before or after having produced a block.

However, delaying the payment of these rewards on the protocol level complicates the system immensely. The unclaimed rewards must be kept somewhere so that total currency amount is predictable. Moreover, it forces us to implement an additional transaction type (claim rewards), which will have to be included in some future blocks. Also the information about who is entitled to claim how much must be stored somewhere. It cannot be computed from history unless we set a time limit after which unclaimed rewards are considered forefeit. On the other  hand, implementing it in a zkApp seems relatively straightforward (all the information above will be stored in zkapp's state, probably even off-chain). So if we deem this feature really valuable, it's a strong argument for a zkapp solution.

jrwashburn:
> it would be reasonable to apply at least the same delay as for delegations (epoch+1). If there is need for signalling, it may be increased. Also, it's worth considering adding constrains (e.g. max 2x increase, or max 50% increase)

@michal0mina I agree these provisions mitigate the concern - and since monitoring by the delegator is necessary in every solution, as long as they would have a reasonable amount of time to re-delegate before the change, this overcomes my objections. I think Epoch +2 is necessary since the Epoch+1 staking ledger would already be baked.

0xstardust:
This thread seems very comprehensive. Interestingly, this decision has some downstream effects. At Kintsu we see a major overlap between this thread and the [liquid staking RFC](https://github.com/MinaFoundation/Core-Grants/pull/23) 

In terms of making a scalable and secure liquid staking protocol, delegation rewards should be permissionless and trustless, and ideally, there is a simple way to claim rewards on-chain. Alternatively, there can be some runtime call made arbitrarily that allows rewards to be claimed.

michal0mina:
Update:

1. I removed issues related to concurrency. It's too close to implementation and likely solved by actions. 
2. Added important challenge/unknown with in-protocol solution: it may affect blockchain tip circuit 

We also discussed deeper zkApp designed that would worked based on epoch's staking ledger.
@mitschabaude would you be able to summarize the design and potential issues/blockers?

cc: @es92

mitschabaude:
> We also discussed deeper zkApp designed that would worked based on epoch's staking ledger. @mitschabaude would you be able to summarize the design and potential issues/blockers?

In our discussion in March, we thought that we can use the `stakingEpochData.ledger.hash` precondition to prove the statement "I am allowed to withdraw staking reward" against the relevant ledger snapshot.

From the ledger hash, we should be able to prove both our account balance and our `delegate` at the time of the snapshot.

However, to prove how much we are allowed to withdraw depends on the _total amount staked to the delegate_. This is where we ended up getting blocked: The total stake per delegate is not something contained in the ledger or succinctly provable against it.

To explore this direction further, I would consider a sort of "zk coprocessor" to Mina which provides this missing information: A map from delegates -> total stake.
This map needs to be updated with every single transaction, so the coprocessor has to process all Mina transactions, and anchor them against the actual ledger, and update a recursive proof which tracks the delegate -> total stake map.

This coprocessor should probably be a shared piece of infrastructure because it is expensive to operate (needs to constantly churn out proofs).

Another option could be to update the delegate -> total stake map as part of the existing protocol, in the transaction snark. That would be a HF-requiring change, but otherwise I think a fairly non-intrusive protocol change.

It also might be the case that "historical preconditions" (which are on the protocol roadmap) are needed to get the relevant ledger snapshot (I don't know enough about staking/consensus to tell whether the existing `stakingEpochData` is the correct snapshot and whether it's realistic that all staking rewards get collected before `stakingEpochData` is updated to a different value)

michal0mina:
@mitschabaude can zk cooprocessor be replaced with an oracle that serves appropriate ledger? Then the contract would be able to compute total delegated stake per delegate? 

I am assuming here once per epoch computation and once per epoch claim. Probably with the zk cooprocessor it would be possible to jump in and out on demand and do not wait full epoch for accounting.

mitschabaude:
> @mitschabaude can zk cooprocessor be replaced with an oracle that serves appropriate ledger? Then the contract would be able to compute total delegated stake per delegate?

The problem is that a computation based on the current ledger would have to process _every single account_ in the ledger and fully re-derive its hash. Otherwise, there is no way to prove that your calculation includes _all_ accounts that are delegating to a particular delegate.

I think processing the entire ledger in a zk circuit, every epoch, might be prohibitively expensive. Starting from a known correct delegate->total_stake map and updating it by using all new transactions applied to the ledger in between might be cheaper.

michal0mina:
@mitschabaude is there any middle ground? Can we use the ledger and hash as arguments and have outside tool to show these are complete and match each other?

mitschabaude:
> @mitschabaude is there any middle ground? Can we use the ledger and hash as arguments and have outside tool to show these are complete and match each other?

that's kinda where my zk coprocessor idea was going at

alternatively you could have a trusted source that signs the data that the zkapp needs. (that source would have to be _very_ trusted though because it determines who gets paid staking rewards)

andrewferrone:
Hey @kantp flagging this old discussion ^

PR #10, opened 2023-12-26
RFC-0006: Review and Redesign of Mina's Tokenomics post Hardfork

lampardlamps:
this RFC requests that Mina Foundation commission a reputable research team to redesign Mina tokenomics to best incentivise protocol developments.


RFCs/RFC-0006: Redesign of Mina tokenomics.md
```diff
@@ -0,0 +1,79 @@
...
... launch (see Section 6. Initial distribution Categories, 1d)Supercharged rewards of [this blog post](https://minaprotocol.com/blog/mina-token-distribution-and-supply)). The winding down of supercharged rewards did not get implemented, though, as the hardfork of smart contract was delayed by almost two years, and the inflation has been continuously running at >20% the whole time. This hyper inflation has been a concern for the community. Even though an MIP has been passed to remove supercharged rewards at the planned hardfork, which is expected in 2024, by the time the unplanned inflation will have run more than 15 months, and the theoretical inflation will still be 8.8% ([calculations by Gareth](https://forums.minaprotocol.com/t/mip-to-review-and-redesign-mina-tokenomics/6131/6): max of 7140 slots and 75% fill rate so 5355 max blocks. 5355 x 720 = 3.8556 million MINA increase an epoch. The current supply is 1.0724 billion, so that’s an increase of 0.00359517041% an epoch, annualized that’s 8.8%. If considering an active stake of 77.71%, the inflation rate would be ~8.0%). In comparison, the inflation rates of major L1 chains, for example, Ethereum, Solana and Cardano, are all < 7%.
```
garethtdavies, line 17: I think hyperinflation is a little dramatic ;) It's certainly higher than it needs to be. Also, staking rate != inflation rate. The inflation rate has historically been around 12%, which is actually very close to the original tokenomics (not really by design, but a combination of fixed coinbase and not all MINA staked).

lampardlamps: Hi Gareth, thanks for your comments. I've modified the wording to 'unplanned high inflation' referring to over-run SC rewards, and added a note on the historical inflation figure. Cheers.

es92:
Just thinking about the structure of such a report - one section that might make sense is for it to include is a review of other protocols with similar economics (block rewards, transaction fees, etc) and what parameters they have selected.

One idea for structuring the report (feedback would be great if this is the right information):
1. A review of the economic parameters / systems within Mina
2. A review of other protocols and how they have setup their token-economics
3. A recommendation of a range of parameters / different options for Mina to consider
4. A discussion how Mina might choose between different options and parameter selections for different preferences, and recommendations of particular parameter selections given different preferences.

We might want to bake into the grant some review rounds as well with opportunities for MF to organize community comments for whoever is doing the report to respond with more information.

In general I think this is good though, and supportive in general of a process for reviewing / discussing where to go with this!

lampardlamps:
> Just thinking about the structure of such a report - one section that might make sense is for it to include is a review of other protocols with similar economics (block rewards, transaction fees, etc) and what parameters they have selected.
> 
> One idea for structuring the report (feedback would be great if this is the right information):
> 
> 1. A review of the economic parameters / systems within Mina
> 2. A review of other protocols and how they have setup their token-economics
> 3. A recommendation of a range of parameters / different options for Mina to consider
> 4. A discussion how Mina might choose between different options and parameter selections for different preferences, and recommendations of particular parameter selections given different preferences.
> 
> We might want to bake into the grant some review rounds as well with opportunities for MF to organize community comments for whoever is doing the report to respond with more information.
> 
> In general I think this is good though, and supportive in general of a process for reviewing / discussing where to go with this!

Hi Evan,
  Sorry for the late response, and thanks for the thoughtful comments - and your support! I've added the recommended structuring into the RFC. I do wonder what the next step for the RFC is - will someone at MF turn this into an RFP which will start solicit the commissioning process? Is there anything else I should do please? - cc-ing @remigiusz-antczak as well. Please kindly let me know. Thanks again.

lampardlamps:
A similar tokenomics optimisation work (on reward mechanism) was published by Gauntlet for Immutable today: https://imx.community/blog/gauntlet-analysis-optimizing-imx-reward-mechanisms

lampardlamps:
Now that the hard fork has happened and zkApps will soon hit the mainnet, it seems inreasingly likely that most activities will take place on L2s, with Zeko or Protokit being the leading framework. How will the Mina token capture the value? We could have a prosperous ecosystem on L2, but those values will likely accrue to their own tokens, like OP or ARB on ethereum. How can we make sure that Mina as the settlement layer is not forgotten? These are questions that need to be explored in the tokenomics design.

PR #11, opened 2024-01-08
RFC-0007: zkPassport

es92:


RFCs/rfc-0007-zkPassport.md
```diff
@@ -0,0 +1,106 @@
...
+| **Requirements** | An implementation of zkPassport in the Attestation API Standard (see objectives), plus an on-chain nullifier that users can use with the zkApp. |
+| **Expected Outcome** | Via leveraging a simple library, users can make zkApps that are sybil-resistant. |
+| **Impact Analysis** | This would unlock many applications, such as one-person-one-vote, one-person-one account, Sybil-Resistant airdrops, etc. |
+
+### Example Scenario 2: Proof-of-nationality
+
+| Aspect           | Description |
+|------------------|-------------|
+| **Description**  | A zkApp wants a user to prove facts about what country they are / are not from. |
+| **Requirements** | An implementation of zkPassport in the Attestation API Standard (see objectives). |
```
teddyjfpender, line 80: Another requirement would be to use the Attestation API to recursively prove many properties of a credential. For example `Prove(Prove(nationality===UK) & Prove(documentExpiryDate>currentBlockHeight))`.

RFCs/rfc-0007-zkPassport.md
```diff
@@ -0,0 +1,106 @@
...
+Web2 applications may find use in this functionality as well. This seems particularly relevant as AI systems become increasingly powerful, and the risk this could pose to new and existing digital systems being overwhelmed by fake users.
+
+## Scenarios and Use Cases
+
+### Example Scenario 1: Proof-of-Uniqueness
+
+| Aspect           | Description |
+|------------------|-------------|
+| **Description**  | A zkApp wants a user to prove they are unique. |
+| **Requirements** | An implementation of zkPassport in the Attestation API Standard (see objectives), plus an on-chain nullifier that users can use with the zkApp. |
```
teddyjfpender, line 71: As well as nullifiers, users can also decouple an on-chain identity from the zkPassport credential with a Diffie-Hellman like key-exchange with the credential issue using viewing keys where viewing key is something like `Poseidon.hash(PrivateKey.toFields().toBigInt())`.

RFCs/rfc-0007-zkPassport.md
```diff
@@ -0,0 +1,106 @@
...
+
+It seems countries keep databases of lost, stolen, and revoked travel documents[[5](https://www.icao.int/publications/Documents/9303_p2_cons_en.pdf)]. It doesn't seem like this is public however, so there may also have to be logic for invalidating one's old passport if one gets a new passport (due to the old one either being lost, stolen, revoked, or just expired).
+
+Lastly, a passport isn't something that everyone has or is able to get, so this is unlikely to work as a universal solution for proving personhood. That said, it could be a useful "anchor", for decentralized social-graph based approaches, which I think could be promising future work.
+
+This could be done via passport-holders verifying non-passport-holders. One way this could work is requiring multiple passport-holders, say 3, to attest to the validity of a non-passport-holder, and limiting the number of times they can do this (if 3 passport-holders are required, and each passport-holder gets 3 verifications, this would mean each passport-holder could verify up to one non-passport-holder on average).
```
teddyjfpender, line 92: Social identity is really cool, I like this one! 💡

teddyjfpender (changes requested):
This is great @es92! Just a couple of comments on format & some potential additional requirements that came to mind 😃

mrmr1993 (commented):
> A remaining question is exactly what new cryptography primitives would need to be implemented in o1js, if any new ones are needed. It looks like RSA, DSA, ECDSA, and Sha-xxx are used (see pages 15, 16, and 17 here[[4](https://www.icao.int/publications/Documents/9303_p12_cons_en.pdf)]).

I've looked around in some other places for confirmation, and this seems to be the full list!

All the tech checks out by my reading, this will be a very exciting addition to the ecosystem.

es92:
> > A remaining question is exactly what new cryptography primitives would need to be implemented in o1js, if any new ones are needed. It looks like RSA, DSA, ECDSA, and Sha-xxx are used (see pages 15, 16, and 17 here[[4](https://www.icao.int/publications/Documents/9303_p12_cons_en.pdf)]).
> 
> I've looked around in some other places for confirmation, and this seems to be the full list!
> 
> All the tech checks out by my reading, this will be a very exciting addition to the ecosystem.

Nice! Do you know if this is already all in o1js after the recent ECDSA/SHA256 work, or is there more that would need to be in any grant on the cryptography front?

RFCs/rfc-0007-zkPassport.md
mazito: I think it will be good to add a link to the mentioned MINA Attestation API.

mrmr1993:
> Do you know if this is already all in o1js

I don't know about the status of the Sha-XXX hashes, but let me ask around. We already have ECDSA, and I've heard about since experiments on the DSA and RSA fronts though, so we're close already!

mitschabaude:
Just came here to say that this is a fantastic idea that seems technically feasible and extremely promising!

teddyjfpender:
Now that the [Attestation API RFC](https://github.com/MinaFoundation/Core-Grants/pull/14) is out, we can link the two together!

RFCs/rfc-0007-zkPassport.md
es92: Oh interesting - I'm not sure how this would work exactly though? What exactly is it accomplishing?

RFCs/rfc-0007-zkPassport.md
mazito: This is interesting but really difficult to avoid collusion of passport holders to validate a fake/malicious non passport holder. The validated malicious non-holder can latter perform malicious acttions without any difficulty. The non-holder may not even exist and be just a way of the 3 holders to act together and perform some action. Having no traceability back from the initial holder approvers to the fake non-holder also makes it quite worse. 

Validating identity by social methods is much more complex. ProofOfHumanity has some way of doing it, we used anonymous trusted validators in the Identicon protocol (the idea that originated Socialcap), and there are others.

mazito:
I think this is quite useful and many apps in MINA may benefit from it. Having a well defined and standard way of doing identity verification is key in many uses cases. Socialcap is one of them, and any credential system may benefit from it also.
In many cases you don't even need to access any personal info, just the verification that this is a "unique" person.

mazito:
Continued thinking about this, and what this will actually prove is that "someone has a given NFC enabled passport in his/her hands".

It does not prove that that person "holding" the passport in his/her hands is the real owner of the passport. 

That is why all KYC programs require you to take a photo of the passport and a real time video/photo selfie to assert that both are the same.

mazito:
And just to finish here (sorry for the long comments) this graph from Vitalik's article helps to define the scope of the NFC passport proposal:

image

I would say that it falls in the "Specialized-hardware biometric" category, because in person biometrics are used and verified by the country issuing the passport when the passport is requested. In fact the passport is a readable paper/electronic proof of that biometrics verification.

es92:
Thanks @mazito , some thoughts:

- Agree on social identity - I think it would be more future work extending this to use some kind of social graph (that would have to be well more advanced than the one here.... but seems plausible and hopefully passports lead to a good starting point)
- Agree on this being proof of holding a passport too - whoever is implementing a production version of this would probably want to take this into account, perhaps by enabling users to deprecate their old passport if they get a new one - I worry that the including a photo / video stream may not work in the future too with generative AI video getting better, so just having a path for users to replace their old credentials seems potentially more robust. It seems likely there would be other considerations too in practice for a production version.
- Is specialized hardware things like worldcoin though? For that I'd agree, maybe this is slightly different though, since anyone can get a passport and scan it? I'd argue that its Privacy High, Accessibility / Scalability High, Robustness of decentralization Fairly Low (depending how much you trust governments issuing passports.... which potentially could be better than a company-run solution), and security against "fake people" probably medium, unless replacing your passport is considered easy than High... Maybe there would need to be a new column, for something like "Govt issued digital ID"?

mazito:
> * Agree on social identity - I think it would be more future work extending this to use some kind of social graph (that would have to be well more advanced than the one here.... but seems plausible and hopefully passports lead to a good starting point)

- Totally agree on this. It is  GREAT starting point. In fact, today the only "trustable" (mostly ...) sources of real identity are government issued documents (passports, identity cards, ...).

- And the fact that it can become a MINA standard is worthwhile too, allowing other apps to leverage it.

> * Agree on this being proof of holding a passport too - whoever is implementing a production version of this would probably want to take this into account, perhaps by enabling users to deprecate their old passport if they get a new one - I worry that the including a photo / video stream may not work in the future too with generative AI video getting better, so just having a path for users to replace their old credentials seems potentially more robust. It seems likely there would be other considerations too in practice for a production version.

- Replacing or renewing passports is a need because also passports have a limited duration (5 years in some countries, maybe diffrent in many others). So a due date for the attested passport should also be considered. 

- Also the option of revocation (for some reason such as someone detecting a fake passport, which can create a request to evaluate it and eventually revoke it). And this is in contrast with the desired "no censorship" feature in blockchains. So this is a complicated issue too :-) 

- Fully agree that photo/videos will be constantly chllenged by the generative AI video getting better. I only mentioned it because most KYC systems use it as a complement to the passport info. May be a weak check, and as you said may not be very useful. BUT in some way you have to check liveness (or proof of life) for the one creating the proff with his/her passport. No good solutions here for now :-(

> * Is specialized hardware things like worldcoin though? For that I'd agree, maybe this is slightly different though, since anyone can get a passport and scan it? I'd argue that its Privacy High, Accessibility / Scalability High, Robustness of decentralization Fairly Low (depending how much you trust governments issuing passports.... which potentially could be better than a company-run solution), and security against "fake people" probably medium, unless replacing your passport is considered easy than High... Maybe there would need to be a new column, for something like "Govt issued digital ID"?

- Yes. There should be another column for "Govt issued digital IDs" (and many govts are doing work on this area). I just mentioned it in the "specialized" column because govt organizations that issue the passports do some sort of "biometrics collection and verification" (taking fingerprints, taking your photo, interviewing you, checkin other existent previous docs, etc). 

- And of course the categories must be reviewed on light of this new proposal. I think security is probably "high" if we can check in some way that the one doing the proof is the same as the passport owner. I think frauds can happen in that particular moment.

- Totally agree with Privacy High, Accessibility / Scalability High, Robustness of decentralization Fairly Low (depending how much you trust governments issuing passports.... which potentially could be better than a company-run solution" 

- Anyway, Govts will continue to issue passports for a long time (and other identity docs too), and as you mentioned "potentially could be better than a company-run solution" (discarding politics here !)

RFCs/rfc-0007-zkPassport.md
Pfed-prog: BrightID is another viable identity model
-------------------------------------
Revoking identity and multiple validations could be another way to solve this

Pfed-prog:
> Also the option of revocation (for some reason such as someone detecting a fake passport, which can create a request to evaluate it and eventually revoke it). And this is in contrast with the desired "no censorship" feature in blockchains. So this is a complicated issue too :-)

What if there are mutliple issuers attesting the same information, similar to having multiple passports.

Verifiable Credentials Data Model v1.1
image

RFCs/rfc-0007-zkPassport.md
```diff
@@ -0,0 +1,106 @@
...
+* placeOfBirth
+* residenceAddress
+* phoneNumber
+
+## Objectives
+
+We propose a grant taking advantage of this, implementing, at a high level, the following:
+
+1. An iOS & Android App that uses NFC to allow users to get the data from their own passports, and store that data in their MINA wallet for secure use by web3 applications through the (in-progress) Mina Attestation API Standard.
+2. An o1js library that uses the Mina Attestation API Standard to create proofs of various components of one's zkPassport identity, for use in proof-of-unique-user and proof of nationality & age use cases.
```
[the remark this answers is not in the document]
Pfed-prog: Very curious if there is a need for a iOS & Android App when NFC could be connected with a standard browser 

Tested the following website with my cheapish Android phone on Chrome

https://whatwebcando.today/nfc.html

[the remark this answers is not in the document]
RaidasGrisk: Using NFC scanning directly in the browser instead of creating separate iOS & Android app would make the process more modular and user-friendly, I think. It would be simpler and more direct: go to a website, scan your passport using NFC, and then transfer the data to your wallet, zkApp, etc.

I think the Web NFC API currently only supports basic scanning functions. With a passport NFC scan, there's an additional step to decrypt the data using a key found on the passport's first page. This particular interaction isn't included in the native Web NFC API, I think. 

I hope this is achievable. Such a fantastic use case.

RFCs/rfc-0007-zkPassport.md
EmrePiconbello: Proof of uniqueness can be a solution here if user don't have passport or prefer not to share the data. Any kind of social system is always open to be gamed

EmrePiconbello:
I just want to go ever few points and share our knowledge on certain aspects of the topic here.

1)There is no way to verify a NFC actually real on it's own. It can be easily faked. That's why whole online KYC/AML process rely on face to to match the OCR extracted data from the passport. 
2)To tackle this issue there are few ML/AI based software and few new upcoming challengers which specialized on these kind of documents and humans face to be really on present at that time. There is few reputable 3rd party auditors for these software. For example ibeta.com which does all test from accuracy, penetration etc. from face scan to fingerprint scanner which governments use in border controls and put in to passport. 
3)Digital ID space booming in recent few years and a lot of governments looking in to ways to do that experiment with solutions and they are very well on privacy side. I can give few examples to that from UK digitalidconnect.com is a standard framework for digital ID. While we have examples like oesterreich.gv.at which is direct digital ID from government which taking the plastic ones out of use. I believe in the next 5-10 years we will have full transition towards digital ID which will be much easier to integrate for ZKpassort. 
4)For special hardware biometrics only cheap solution would be fingerprint scanners (To my knowledge there is only 1 company which creates portable/small fingerprint scanner usable in this way)which still could cost 40-100$ + logistical challenges/cost of distributing all those scanners. 
5)We believe the biggest issue is duplication of identities. Because space is growing very fast everyone is trying different stuff. For example digitalidconnect is a standard for just UK. Even if we have global standards set and stone. I don't think the isolated structure of these won't change because all these solutions access to specialized data from local government database etc. to verify and I don't see governments sharing all this kind of data in some global system would be sensible.

mitschabaude:
> There is no way to verify an NFC actually being real on its own. It can be easily faked

@EmrePiconbello according to the RFC, passport data is signed by the host country. we would verify that signature in the proof. you wouldn't be able to fake that.

EmrePiconbello:
> @EmrePiconbello according to the RFC, passport data is signed by the host country. we would verify that signature in the proof. you wouldn't be able to fake that.

@mitschabaude I am not saying that NFC data is forged. Issue is NFC is very easily cloneable and the verification signature is part of it. While you can't alter the data(data can be outdated etc. those are still the downsides). On it's own nfc part of very easily duplicated. You just need to close enough with nfc device for few second. Implementing that would be very wrong in my opinion just because of that reason.

RaidasGrisk:
> NFC is very easily cloneable and the verification signature is part of it.

Besides reading the data from the NFC chip, seems like it is also possible to verify if the data and chip are genuine and not cloned or faked.

https://www.inverid.com/blog/cloning-detection-epassports
https://secureidentityalliance.org/ressources/blog/secure-chips-trust-in-passports-what-is-pki

es92:
> > @EmrePiconbello according to the RFC, passport data is signed by the host country. we would verify that signature in the proof. you wouldn't be able to fake that.
> 
> @mitschabaude I am not saying that NFC data is forged. Issue is NFC is very easily cloneable and the verification signature is part of it. While you can't alter the data(data can be outdated etc. those are still the downsides). On it's own nfc part of very easily duplicated. You just need to close enough with nfc device for few second. Implementing that would be very wrong in my opinion just because of that reason.

Thanks on this - just added a section commenting on exactly what is being proved here and on security.

EmrePiconbello:
> > NFC is very easily cloneable and the verification signature is part of it.
> 
> Besides reading the data from the NFC chip, seems like it is also possible to verify if the data and chip are genuine and not cloned or faked.
> 
> https://www.inverid.com/blog/cloning-detection-epassports https://secureidentityalliance.org/ressources/blog/secure-chips-trust-in-passports-what-is-pki

We did research on this over a year. Unfortunately I don't have like resources on me right now but with quick google I find this. 
https://eprint.iacr.org/2005/095.pdf There was a more recent paper with a lot of details but I couldn't find it but I have this https://whenderson.dev/blog/biometric-passports/ The implementation and standards are still very similar. There are few new standards coming up with active writable chips but they are still far from being utilized and biggest issues is comes from that since implementations are all around the place even though there is a standard.

I like to summarize what I know concisely. We have many methods. Chip Authentication, is secure can still be compromised if a chip is extracted from a legitimate passport and placed into a fake one not to mention it doesn't protect from cloning. However, integrating the chip brings complexity, and many countries, including the majority of the EU, do not utilize it. Not to mention chip doesn't block reading the data. Since there is no standard and it's all around the place with suggested not mandatory implementations. Even when better standard is implemented support for older ones existing just makes it insecure. 

Additionally, we have a system that verifies these different types of authentication systems. However, due to the lack of uniformity and insufficient security measures behind these methods, as evidenced by extensive research with companies developing hardware and software, along with third-party auditor firms testing these systems, we have concluded that matching all elements provides the highest level of certainty the person and document present at that time. This matching aligns with legal standards for verification. Our focus remains on biometric authentication solutions, as there are few products we wish to explore in this area. We are currently looking in to portable fingerprint scanners viability.

While we are not experts in this space, considering the global usage of the same few software and hardware solutions utilized in law enforcement, border controls, and banking, despite differing approaches, instills confidence.

RFCs/rfc-0007-zkPassport.md
```diff
@@ -0,0 +1,106 @@
...
+| **Description**  | A zkApp wants a user to prove facts about what country they are / are not from. |
+| **Requirements** | An implementation of zkPassport in the Attestation API Standard (see objectives). |
+| **Expected Outcome** | Via leveraging a simple library, zkApps can require a user to prove facts about their nationality. |
+| **Impact Analysis** | This would be useful across web3, to make applications that are regulatory-compliant. |
+
+## Open Issues and Discussion Points
+
+A remaining question is exactly what new cryptography primitives would need to be implemented in o1js, if any new ones are needed. It looks like RSA, DSA, ECDSA, and Sha-xxx are used (see pages 15, 16, and 17 here[[4](https://www.icao.int/publications/Documents/9303_p12_cons_en.pdf)]).
+
+It seems countries keep databases of lost, stolen, and revoked travel documents[[5](https://www.icao.int/publications/Documents/9303_p2_cons_en.pdf)]. It doesn't seem like this is public however, so there may also have to be logic for invalidating one's old passport if one gets a new passport (due to the old one either being lost, stolen, revoked, or just expired).
```
EmrePiconbello, line 88: They have their own databases but it's not public. When passport is issued. The certification and data of it is valid on average 10years +3 months. So if they are stolen, lost etc. You can't use it to exit the country since they would know while you trying to leave. To my knowledge from NFC stand point there is no difference and I don't know any country with public DB on these documents.

RFCs/rfc-0007-zkPassport.md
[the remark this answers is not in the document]
0xturboblitz: Yep, the Web NFC API doesn't support reading passport unfortunately.

lampardlamps:
@es92 cc @teddyjfpender and @mitschabaude there have been quite a few really good zkp ID proposals in zkIgnite 3, see, for examples: 

https://zkignite.minaprotocol.com/zkignite/zkapp-cohort-3/feedbackandrefinement/suggestion/652
https://zkignite.minaprotocol.com/zkignite/zkapp-cohort-3/feedbackandrefinement/suggestion/739
https://zkignite.minaprotocol.com/zkignite/zkapp-cohort-3/feedbackandrefinement/suggestion/741

Each of them have sound and unique implementation plans, but they all touch upon the interactions with passport. I'm sure etonec is similar too. 

To fully take advantages of the talent and make them pull in the right direction, I personally think it would be very beneficial for Mina Foundation to form a working group on zkp ID, regardless of the funding outcomes of the zkIgnite proposals, so that these teams can participate in the discussion about zkPassport and expand its horizon on potential applications.

RaidasGrisk:
> @es92 cc @teddyjfpender and @mitschabaude there have been quite a few really good zkp ID proposals in zkIgnite 3, see, for examples:
> 
> https://zkignite.minaprotocol.com/zkignite/zkapp-cohort-3/feedbackandrefinement/suggestion/652 https://zkignite.minaprotocol.com/zkignite/zkapp-cohort-3/feedbackandrefinement/suggestion/739 https://zkignite.minaprotocol.com/zkignite/zkapp-cohort-3/feedbackandrefinement/suggestion/741
> 
> Each of them have sound and unique implementation plans, but they all touch upon the interactions with passport. I'm sure etonec is similar too.
> 
> To fully take advantages of the talent and make them pull in the same direction, I personally think it might be very beneficial if Mina Foundation could form a working group on zkp ID, regardless of the funding outcomes of the zkIgnite proposals, so that these teams can participate in the discussion about zkPassport and expand its horizon on potential applications.

Just to add to the list, we've developed a fully functional zkp ID app as part of zkIgnite 2, called [**id-mask**](https://idmask.xyz/). Currently, the app's reach is limited geographically because we've only integrated it with a single local KYC personal data provider. Eager to see how can we leverage passport-NFC interaction.

Moreover, I believe there are some unexplored issues that every zkp ID app will encounter later on. We briefly discussed this with @EmrePiconbello on Telegram. How can we design systems that prevent others from sharing proofs? For instance, once a zk-proof is created using passport-NFC, how do we ensure that only the original creator of the zk-proof can utilize it, preventing unauthorized sharing?

RFCs/rfc-0007-zkPassport.md
```diff
@@ -0,0 +1,150 @@
...
+Web2 applications may find use in this functionality as well. This seems particularly relevant as AI systems become increasingly powerful, and the risk this could pose to new and existing digital systems being overwhelmed by fake users.
+
+## Scenarios and Use Cases
+
+### Example Scenario 1: Proof-of-Uniqueness
+
+| Aspect           | Description |
+|------------------|-------------|
+| **Description**  | A zkApp wants a user to prove they are unique. |
+| **Requirements** | An implementation of zkPassport in the Attestation API Standard (see objectives), plus an on-chain nullifier that users can use with the zkApp. zkPassport nullifier proofs should be sure to have a *context* input salt that is hashed with the unique zkPassport identifier to ensure identities are unique across their context, but cannot be associated cross-context unless a proof is provided by the user. |
```
es92, line 71: @KimlikDAO-bot this is one way to ensure that identities aren't traceable cross application - the salt can make sure that how one proves uniqueness for one application isn't traceable across applications.

It would probably be important that however wallets interact with nullifiers, wallets check and make sure the same nullifier-key isn't being used across applications

EmrePiconbello:
Thanks for the suggestion @lampardlamps on that note I like to add this.

Because we have been conducting our research and planning for a very long time, I mentioned something similar to this in our proposal since zkIgnite 1 or 2. Now that we are preparing to launch the product, I want to discuss the details and establish some kind of standard, as our aim is to link all ID solutions to a single ID. Recently, @RaidasGrisk was very helpful with his insights on the matter. He developed ID-Mask, and I believe we have found a middle ground with a structure like hashed (name, surname, personal number) as public output. There is a lot of variance, which could be a limitation for some solutions, as not all solutions might include a personal number. I firmly believe we need these discussions to establish some kind of standard/best practice so that future interactions between these zkApps would be smoother. Otherwise, we will end up with many solutions that are disconnected from each other or cannot work with other zkApps.

es92:
@RaidasGrisk 

> Moreover, I believe there are some unexplored issues that every zkp ID app will encounter later on. We briefly discussed this with @EmrePiconbello on Telegram. How can we design systems that prevent others from sharing proofs? For instance, once a zk-proof is created using passport-NFC, how do we ensure that only the original creator of the zk-proof can utilize it, preventing unauthorized sharing?

This should be covered by the Attestation API, to make sure it doesn't leave the wallet of the user who has created the attestation. This would ensure browser pages cannot access the original credential, and only get context-specific proofs.

es92:
Agree & that definiteley makes sense re some kind of working group btw, I can check in if there's bandwidth to organize it or if it would need to be more adhoc, lms

RFCs/rfc-0007-zkPassport.md
es92: Yeah that's a good point. I think for this passport case, I think there should be enough data to avoid this, by using the passport signature from the authority itself as part of the hash - so `unique_passport_hash = hash(passportSignature, passportData)`, and it shouldn't be brute-forceable I think.

And then using `hash(application_salt, unique_passport_hash)` for any nullifiers on-chain to avoid revealing the unique_passport_hash, seems like it should work. Even if the application_salt is constant, there shouldn't be a way to find the unique_passport_hash then, (or the additional requirement of creating a proof that one possesses the data underlying that hash that has been signed by the passport authority).

Any group implementing / auditors should definitely make sure this works out in the implementation and specifics of any passport cryptography though, as well as put together a more complete spec. Curious if you think this seems right or any other thoughts though.

es92: Ah I see, yeah that makes a lot of sense. Yeah it would probably be best to avoid the passport authority being able to do that.

Maybe the following could be done to do this without needing a multiparty protocol? (and any additional trust assumptions that would need to follow from that?)

1) have an "identity-registration" zkApp, that uses a nullifier using the passport-authority-visible hash as the key, to allow users to register their own user-selected unique hash.
2) have other applications use the user-selected hash for proofs of uniqueness - which should work because each user-selected hash corresponds to an passport-authority-generated hash, and there isn't a way to generate multiple user-selected hashes per passport-authority generated credential.

Curious on any thoughts? Seems like something multistage is needed though yeah for on-chain privacy from the passport authority.

es92: Interesting re the encryption size thing - I'd think that if its big enough to not be brute-forceable to fake-sign passports, it should be enough entropy for not brute-forcing a hash of all of the data? Though I haven't looked at the cryptography in detail, and would definitely be curious for a cryptographers take.

EmrePiconbello: The encryption is there to protect the passport data but majority of implementations are very week at that level. The certificate signed by issuing government is mostly safe. There are few mistakes done by few governments in the history while transitioning towards the e-passports but I don't expect similar incidents being a issue in the future.

es92: I'm not sure, it should be sufficient I think to get this property (on-chain privacy from the issuer) through an on-chain operation? Could totally be missing something though.

In particular:
1) Alice generates S, their private, unique identifier, and stores it off-chain
2) Alice receives P, their unique passport credential, and H(P), the hash of P, which is private to Alice and the passport-issuing-authority.
3) An on-chain identity zkApp is created with a nullifier for registering identities, which accepts a proof, key, and value, and application salt A.
4) Alice posts to the on-chain identity zkApp:
    a) a proof that Alice posseses P signed by the passport issuing authority
    b) a key into the nullifier, H(A,H(P))
    c)  a value to register into the nullifier, H(S)
5) Afterwards, Alice proves to other applications that she knows of an S that has an H(S) in the on-chain identity zkApp nullifier, and when using other applications, nullifiers, uses the key H(Application_key), S).

Because S is private to Alice, nobody else can know which passport-authority-issued identity her identifier S is tied to, or her private value S, so it should be resolve potential privacy issues with the passport-issuer, but lmk, definitely could have missed something.

EmrePiconbello: I believe this is mostly inline with what we were discussing, with @RaidasGrisk.

mitschabaude: @es92 if the key is a Merkle map key (that's how I understood your idea), then you have to share that key as you update the onchain Merkle root. Otherwise, others can not keep track of the updated Merkle tree

mitschabaude: What would be nice is some sort of non inclusion proof into a set without revealing anything else

mitschabaude: Maybe something like a bloom filter, where the hash of your unique secret is smashed together with others and so by proving that your set addition toggles 1 bit, you only reveal a tiny amount

mitschabaude: Actually, what's so bad about revealing a passport hash at the time of registration in that zkApp. You can do that from any account. If you later prove inclusion you don't reveal anything. I.e. I think @es92 your solution is good.

Also, why does it have to be an application specific hash? Why not take H(P) as the Merkle map key?

Pfed-prog:
> Moreover, I believe there are some unexplored issues that every zkp ID app will encounter later on. We briefly discussed this with @EmrePiconbello on Telegram. How can we design systems that prevent others from sharing proofs? For instance, once a zk-proof is created using passport-NFC, how do we ensure that only the original creator of the zk-proof can utilize it, preventing unauthorized sharing?

You can ensure that the sender is the owner by implementing a function with struct that checks who sent the message and whether the struct owner is the same as the sender

RFCs/rfc-0007-zkPassport.md
Pfed-prog: My understanding is that 10 years is the number due to expiration of NFC tags

Pfed-prog: In Belgium citizens can log in to their government portal with eID card reader
https://www.youtube.com/watch?v=RBFe-FdPw34

RFCs/rfc-0007-zkPassport.md
Pfed-prog: I dont think the solution is to track every single person all the time

RaidasGrisk:
>> This should be covered by the Attestation API, to make sure it doesn't leave the wallet of the user who has created the attestation.

> > You can ensure that the sender is the owner by implementing a function with struct that checks who sent the message and whether the struct owner is the same as the sender
> 

Do these solutions assume that on chain address is linked to single IRL identity?

Let's think about proof-of-adulthood again. Two people work together to cheat the system: one is an adult, the other is underage. The adult uses their passport (or other means of passing private data) to create proof-of-adulthood. After making the proof, they link it to a public address (ignoring how it's done). Then they give the underage person the private key for that address. This tricks the system, right?

Is this a big problem with the system? The way we currently check for adulthood can also be cheated in similar ways. But if we aim for adoption and if more people start using the new system, they'll look for security problems. Would the government approve of this system if it has such flaws? I'm not sure of a good solution, just want to point out the problems so we can find good solutions.

mitschabaude:
> Do these solutions assume that a public address is linked to single IRL identity?

@RaidasGrisk Evan's architecture [outlined here](https://github.com/MinaFoundation/Core-Grants/pull/11#discussion_r1484811875), and also the system of Worldcoin which is similar, ensures that every IRL identity can only be linked to a single public key. (That property is needed to make it a proof of uniqueness!)

So, yes, an adult can create a unique ID and then let their child use their computer (absolutely unavoidable feature of every digital system). But no, they can't create multiple wallets all "proved to be owned by an adult" and then hand them out. So, AFAIU, the issue you're concerned with is avoidable.

Btw, re ensuring in a zkApp that a user owns a certain address: It's easy - you just need to create an account update for that address and require a signature on it.
```ts
let update = AccountUpdate.create(userAddress);
update.requireSignature();
```

@Pfed-prog the solution you posted is insecure, because `this.sender` is not proved to be the actual sender of the transaction:
https://docs.minaprotocol.com/zkapps/o1js-reference/classes/SmartContract#sender

RFCs/rfc-0007-zkPassport.md
mitschabaude: @KimlikDAO-bot I understand that you can brute-force `P`, i.e. the users personal data, from `H(P)`.

However, as I've tried to convey, this shouldn't be a problem, because you wouldn't use `H(P)` as your "ID". Instead, you would only use `H(P)` in a registration step, to register a _different_ ID, `H(S)` - let's say, for simplicity, a hash of your private key.

After that, you would be able to create "proofs of unique personhood" _without revealing anything about your actual identity_: You demonstrate, in a zk proof, that you know _private inputs_ `S` (private key) and a `P` (passport data) such that `H(S)` is included in a Merkle tree, at position `H(P)`. This zk computation would just be part of a zkApp which requires your unique-humanness.

mitschabaude: So in a sense there is no "number that is your unique id". That is not needed. What's needed is a way to zk-prove that you _have_ a unique id

mitschabaude: > Anyone who has read the NFC chip of your passport (airlines, airport authorities) or anyone who can recreate that (government signing authorities) can immediately link your wallet address to your government ID

They can link the _address that you register your id from_ to your personal data. However, your later actions are unlinkable with this step.

This is very similar to the kind of anonymity that protocols like ZCash or Tornado Cash give you: There is a step where you send money into a shielded pool, from a public address. So, everyone can see that this address sends money into the pool. In our system, everyone can see that some person registers for a unique id.

But after registering, your actions are protected by the anonymity set of all the people who also registered.

mitschabaude: However, it feels like front-running is a dangerous issue in that design @es92
Someone who sees that transaction registering `H(P)` into the Merkle map can just register his own secret to the same `H(P)` first, thus stealing your passport identity 😨

mitschabaude: Well actually I think the front-running thing can be solved with a commit-and-reveal scheme. You "preregister" first, by showing a blinding commitment to `P` i.e. `H(S, P)`. And it's only possible to register once n blocks have passed since a preregister was done. So, unless someone manages to prevent you from posting a transaction completely for n blocks, they aren't able to front-run it because they need to create their own preregistration first

RFCs/rfc-0007-zkPassport.md
EmrePiconbello: What you outlined here is eID not a passport. The 10 years is because passports are mostly valid for 10 years the problem is NFC is permanent the certificate is 10 years + few months so if it's stolen or something it will be always valid when you are checking it according to certificate.

Pfed-prog: No, NFC expire in 10 years. I have seen this number on vendors pages, articles, etc

For example, 

> Finally, NFC tags could also be deleted if their data retention time expires. Data retention on NFC tags is intended to last for a set amount of time, typically 10 years. The tag might not be able to keep data properly if the data retention time has passed.

https://taptag.shop/blogs/how-tos/ways-an-nfc-tag-or-nfc-card-can-be-erased-magnets-heat-etc

Pfed-prog: It does not matter if it is a drivers license, residence card, passport or eID NFC tags expire in typically 10 years

Pfed-prog: https://taptag.shop/blogs/how-tos/what-is-the-difference-between-ntag213-ntag215-and-ntag216

> What is the difference between NTAG213, NTAG215, and NTAG216?
> 
> Number of scans: The NTAG213 can endure up to 50,000 scans, whereas the NTAG215 can withstand up to 200,000 scans. 500,000 scan operations are no problem for the NTAG216.
> 
> Operating frequency: The 13.56 MHz frequency is the same for all three tags.
> 
> Read/write distance: Depending on the NFC reader and the surrounding conditions, NTAG213, NTAG215 and NTAG216 all have a read/write distance of up to 4 cm.
> 
> Data retention: The 10-year data retention period applies to NTAG213, NTAG215, and NTAG216.
> 
> Data retention: The 10-year data retention period applies to NTAG213, NTAG215, and NTAG216.

EmrePiconbello: The NFC retention is totally depend on standard it utilize, there are NFC's which can hold data for 20-50 years etc. The topic we are talking here is validity of the document according to read we can get from NFC. The validity of the data is rely on certificate. Issue is when that document is invalid certificate or any data doesn't get updated because it's a passive system. You are just taking the conversation out of focus here in my opinion.

Pfed-prog: Well, my passport has one of these tags, not the one that can hold data for 20-50 years etc

Pfed-prog: I just think its quite extraordinary that we want to take in the account only the best case scenario and disregard all the rest, funny

EmrePiconbello: How long NFC store the data doesn't matter here. The important part is is it valid or enough to verify identity or not. I am saying it's not because certificate stored in NFC chip can be cloned or passport can be stolen and there will be no way you can check for that.

Pfed-prog: Sure, but we can create an api service that helps in emergencies and that stores and is responsible for the data storage

Pfed-prog: Here is NFC Tag Issuance Process from Near Field Communication (NFC) Technology with Jordan Mobile Payment (JoMoPay)
image

RFCs/rfc-0007-zkPassport.md
es92: > Well actually I think the front-running thing can be solved with a commit-and-reveal scheme. You "preregister" first, by showing a blinding commitment to `P` i.e. `H(S, P)`. And it's only possible to register once n blocks have passed since a preregister was done. So, unless someone manages to completely block you from posting a transaction for n blocks, they aren't able to front-run it because they'd need to create their own preregistration _after_ seeing your data

Yep - further, the zkApp can require a proof that alice knows S in H(S), so knowing H(S) shouldn't be enough to front-run then either (if we're talking about front-running H(S)).

I'm not sure on the entropy, but I'll add that to a "todo" section of this RFC - I'd think there'd be enough entropy between all the data in P (includes I think maybe fingerprint / photo data as well?) and the signature from the issuing authority, but its definitely something someone on the cryptography side should check (it sounds like definitely a risk at least to make sure of).

EmrePiconbello:
> @EmrePiconbello raises a valid point regarding cloning. Besides this, people give their passports to others for all sorts of legit reasons (getting a visa, checking into a flight etc). Passports get misplaced or stolen all the time.
> 
> There is another, more subtle and more severe issue in the proof-of-uniqueness proposal outlined here. The proof-of-uniqueness secret has to be computed from some constant value such as the user's government unique ID. Whatever computation the user does in isolation can be replicated by a motivated actor to link all on-chain traces to people's government IDs. Note that even the nullifiers are deterministic functions of the proof-of-uniqueness secret, so the motivated actor can obtain the user's government ID just by looking at the nullifiers. (note government unique ID search space is tiny, even if it weren't guess-and-check attacks make this a no-go)
> 
> Fortunately, one can engage in a trust minimized zero-knowledge protocol, which achieves info-theoretic unlinkability between the user's IRL ID and the proof-of-uniqueness secret. See this (for now; we'll make a public doc as soon as we can) https://zkignite.minaprotocol.com/zkignite/zkapp-cohort-3/feedbackandrefinement/suggestion/739/detail

How do you achieve uniqueness here I don't get it.

RFCs/rfc-0007-zkPassport.md
EmrePiconbello: Depends on passport since they are all around the place but majority of them have very low entropy because of limited bit space they have and data they need to put in to that space.

RFCs/rfc-0007-zkPassport.md
mitschabaude: Ideally, any architecture makes it possible to override a previously registered passport with a newer one. That seems to be an important fallback flow in case your identity is stolen. If that is possible, it doesn't matter how long the original passport or NFC tag is valid

EmrePiconbello: I don't think that covers cloning or stolen at all. As random people from earth's population. I will never learn my identity utilized on mina blockchain in the first place.

RFCs/rfc-0007-zkPassport.md
EmrePiconbello: https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://eprint.iacr.org/2005/095.pdf&ved=2ahUKEwiBx5DusKOEAxVO9gIHHUo8AaYQFnoECDYQAQ&usg=AOvVaw2LKSzfBduBGPNOm2sFL_7F

This is the most detailed document. While this is old if you do your research you can find this still applies to many passports globally. The biggest problem comes from general practices since this paper didn't change. Few countries changed their structure and it's safer, majority of them didn't make that. There is a new standard published but it's a big question when it's adopted or even at all.

RFCs/rfc-0007-zkPassport.md
mitschabaude: I agree, it doesn't cover that (a liveness  test is also needed), but in an ideal system you can still override an old passport + liveness test with a newer one, and so it doesn't matter how long the old one is valid

mitschabaude:
> How do you achieve uniqueness here I don't get it.

I looked at @KimlikDAO-bot's proposal and wanted to highlight it because it shows a way to completely avoid the privacy concerns of registering with a potentially brute-forceable passport hash which has to be sent to the chain in the open.

I'm not fully convinced that the protocol _as written_ in [the zkIgnite proposal](https://zkignite.minaprotocol.com/zkignite/zkapp-cohort-3/feedbackandrefinement/suggestion/739/detail) achieves uniqueness/privacy, but the core idea is very good. I'll describe a slight variation of it here, that I think achieves both. For simplicity I only consider the passport NFC data and no extra liveness check.

The protocol consists of 3 steps, which I first summarize in non-mathematical form:
1. A **pre-registration step** where the user posts their passport data in randomized/blinded form, and also prove that it comes from a real passport data with signature.
2. An **encryption assistance step** where 2 different members of a public committee each "encrypt" the blinded passport data using a private key that only they know. The private keys of this committee have to be _stable_. If a user can re-run the same protocol again with access to a _new_ set of private keys, that would give them a second human id from the same passport.
3. A **final registration** step where the user takes the encrypted passport data and _removes their own randomness_ that they initially added. This yields data which is still _encrypted_, but no longer randomized: It's _deterministically derived_ from the passport. We can call the result a _unique human ID_. Since it combines the encryption by two committee members, neither of them can re-derive the ID from a given passport, and so neither of them can brute-force the data (except if they collude).  
  The entire computation of step 3 is done in a zkapp method and results in successfully storing your wallet address (or a commitment to it) in a Merkle tree of humans, under the unique human ID as key.

This protocol achieves creating a Merkle tree of unique humans, which can now be used in several ways - either for anonymous use cases like private voting, where you just show that you own _one_ of the addresses in that tree but not which one, and post a nullifier (= so that every human can only vote once) -- or even for use cases where you interact with your address in the open and can prove that it is that of a unique human.

The committee really only exists to assist with encryption. They are not trusted with any authority, and they have no access to the private passport data. Only by colluding and putting their secret keys together they might be able to brute-force the original passport data, and learn the identity of a user who registered (but not necessarily the address they registered with if we only store a commitment to that). We can increase the number of committee members from 2 to n to gain more confidence that at least 1 of n members will not collude.  

------------------------------------------------------------------------

## Mathematical details

(this is one possible version, I'm sure there are others. this version only relies on provable operations that exist now in o1js)

In short, the unique human ID that is computed in the final step is
$$
id = H(apG) + H(bpG)
$$
where $G$ is the generator point of the Pallas curve, $H(\cdot)$ is the Poseidon hash function, $p = H(P)$ is a hash of the passport data $P$, and $a$ and $b$ are the stable private keys of the encryption committee.

$id$ as defined above is a deterministic function of $p$, but can't be re-derived from $p$ except by someone who knows _both committee private keys_ $a$ and $b$. 

Note: Pallas is the curve that we can efficient operations on in a circuit.

**Preregistration**. The user calls a zkapp method `preregister()`, which proves that their passport data $P$ is correct, by verifying their government's signature on it. We compute $p = H(P)$ (Poseidon hash), and treat $p$ as a _scalar_ of the Pallas curve . We can easily treat the Pallas base field element $p$ as a scalar, because the scalar field is larger than the base field.

The user also passes in a random scalar $r$ and in `preregister()` we compute $R = r p G$, using one foreign field multiplication $rp$ and one scalar multiplication. The resulting curve point $R$ is posted as an action to the zkapp (or similar).

**Encryption assistance**. This can happen offchain. Recall that the two members of the encryption committee have private keys $a$ and $b$, and assume that their public keys $aG$ and $bG$ are stored as state on the zkapp.

The user sends $R$ (the randomized passport hash) to both members' nodes, and each sends back $R$ multiplied with their private key -- $aR$ and $bR$, respectively. (Note: $R = r p G$ doesn't reveal $p$, because $r$ is a random secret only known to the user; so the committee members learn nothing from this interaction).

**Final registration**. The user calls a `register()` zkapp method, which does the following:
* It proves that there is a preregistration action containing $R = rp G$, and that we know $p$ and $r$.
* Multiply both $aR$ and $bR$ with $(rp)^{-1}$, which yields $aG$ and $bG$ - the committee's public keys. Check that those really equal the public keys stored in the zkapps onchain state. This proves that $aR$ and $bR$ are of the claimed form. Note: this calculation works because
$$
(rp)^{-1} aR = (rp)^{-1} (arp) G = aG
$$
* Multiply $aR$ and $bR$ with $r^{-1}$, to give $apG$ and $bpG$. Compute the unique human id, $id = H(apG) + H(bpG)$.
* Use $id$ as key into a Merkle map to store the user's address (or, if we want to increase privacy in the case the committee is colluding, we could store some blinding commitment to the address).

mitschabaude:
@KimlikDAO-bot 

> The humanID secret is simply the sum of n parts, each of which is a (deterministic) hash of the user government ID number (example: user's SSN for a user from the US). This ensures that it's unique, right?

Ok, so here's how I understood your proposal -- the "shares" are the BLS signatures by the nodes after we multiplied by `1/Poseidon(r)`, so the sum of the shares is something like $aX + bX$ (where $a$ and $b$ are validator private keys, $X$ is the passport hash to a BLS12-381 curve point).
I thought that if we know the public keys of the validators i.e. $aG$ and $bG$, then we can brute-force the value of $X$ since thanks to the pairing we can efficiently test whether

$$
(aX + bX, G) \stackrel{?}{=} (X, aG + bG)
$$

for any candidate $X$. If this equation is true, then we know we have found the right $X$.

This can be easily fixed by hashing the $aX$ etc before using it as a public human id! (This is why I added the hashes $h()$, even though its for a non-pairing curve, just to destroy mathematical structure.

The second issue is that you don't really describe (or maybe I missed it!) how the user proves to _other parties_ that he has a valid, unique human id. You write

> Note these signatures from each color class are interpreted as "verifiable hashes" in that the user can verify that the hash was computed truthfully by performing a signature check with the public key corresponding to each color class. The final HumanIDv2 secret is the sum of all the shares is mod V, the Vesta scalar field size.

> Crucially, the BLS12-381 signature verification does not happen inside a circuit, but happens as normal computation, so this protocol should be quite fast.

However, if I'm concerned about uniqueness, it's surely not enough that _the user_ knows that they have a valid unique id. Others must be convinced of this as well! Otherwise the user could just make up a number and say "this is my unique id".
That's why in my version I do the whole verification in a circuit. But you write "Crucially, the BLS12-381 signature verification does not happen inside a circuit", so that confused me.

Both of those issues are easy to fix and I really appreciate your input and proposal @KimlikDAO-bot!

EmrePiconbello:
I want to get some details about brute forcing argument with hashing since I believe there is enough entropy. Let me start with example.

Leyla Shashi Bláha
Lilianne Matthew Wragge
Nox Charissa Ahmad
Cardea Godehard Farrell
Fastúlfr Sixte Swango


Here are 5 outputs from a random name generator. Considering all languages and different symbols and them being different codes in Unicode. People can have 2 names or even more. There is no standard for GOV ID numbers either, with some having not just numbers, and their lengths varying greatly. With all these variables, I don't understand how this can be brute-forceable since any hash we see can be from any country with any form of different structure. 

Creating this kind of hash's aim was making possible same user can be know to zkapps by different ID providers. If that's not a solution we still need some kind of standard to make this communication possible. So I like hear people ideas on that matter and how we can achieve that.

mitschabaude:
> Considering all languages and different symbols and them being different codes in Unicode. People can have 2 names or even more. There is no standard for GOV ID numbers either, with some having not just numbers, and their lengths varying greatly. With all these variables, I don't understand how this can be brute-forceable since any hash we see can be from any country with any form of different structure

IMO we want to prevent even very common names, and passports that have the most standardized GOV ID, from being brute-forceable.

Also, I think it could be an issue to include the name in the hash that forms a unique ID, since names do change at least once in the life of many people. They could get a second "unique" ID after marrying :D

@EmrePiconbello do you know whether parts of the data on each passport are absolutely guaranteed to be unique over the lifetime of a citizen? GOV ID maybe?

If yes, just this unique part combined with the protocol [outlined above](https://github.com/MinaFoundation/Core-Grants/pull/11#issuecomment-1937877101) could give us a unique ID with very strong properties, which should be highly reusable across applications.

mitschabaude:
> @mitschabaude With your permission, I'll update our proposal to fix all the issues in the protocol and to give you due credit.

Sure, thanks!

> Btw, `preregister()` is not an on-chain thing right? We should be able to compute the humanID completely off-chain, but talking with the helper nodes.

The idea for preregistering was to avoid the problem with getting frontrun, i.e. as you're trying to register your humanID (sending it in the open), someone else does so before you can. I was thinking about the specific context of registering into a Merkle tree of confirmed unique humans, which I thought was useful.

> We need an on-chain tx only when presenting a context specific nullifier of it, e.g., Poseidon(zkapp_address, human_id_secret) along with a proof. I may be missing something though.

At first glance this sounds like it could also work.

Although I see potential downsides with not recording that stuff onchain:

- Keeping the whole context of the humanID creation process, which involves custom secrets or the full zk proof, offchain on the user's side, might require a custom wallet or other custom logistics. If humanID + associated address are recorded onchain as in my design, then a lot of interesting applications already become possible without any custom wallets. For example, there can be a zkapp which offers a simple `assertUniqueHuman(address, appContext): nullifier` method, which every zkapp can just access by `npm install`ing and calling into that method.
- Furthermore, you require that everyone who wants to be convinced of your unique ID always has to verify a zk proof. In particular, other zkapps which want to use it have to recursively verify the proof. This causes them to use a lot of constraints and makes their proofs bigger, which is avoided if verifying your ID is just a Merkle lookup

mitschabaude:
> When generating the human id nullifier, my circuit would prove that there exists a proof of correctness for the humanID and that the nullifier is computed truthfully using the correct human_id.
> 
> The exact details above may be inaccurate, but can't we efficiently prove the existence of a correct proof and then prove other things such as correct nullifier computation?

We can do that, but again, both the proof of correctness of the humanID and especially the recursive proof verification take quite a lot of constraints that make everything that would use humanIDs heavier than necessary

EmrePiconbello:
> IMO we want to prevent even very common names, and passports that have the most standardized GOV ID, from being brute-forceable.

GOV ID is mostly a 11 number. Since there is no standard there might be regions with less and that can be a problem even if we are combining it name and surname prehash.

> Also, I think it could be an issue to include the name in the hash that forms a unique ID, since names do change at least once in the life of many people. They could get a second "unique" ID after marrying :D

Yes it can change. Considering everything like these with about a year of research we come up with this proposal to cover everything. https://zkignite.minaprotocol.com/zkignite/zkapp-cohort-3/feedbackandrefinement/suggestion/652/discussions uniqueness actually comes from some ml extracted data from face being hashed.  
> @EmrePiconbello do you know whether parts of the data on each passport are absolutely guaranteed to be unique over the lifetime of a citizen? GOV ID maybe?

GOV ID doesn't change but not all passports have gov ID in it :)

mitschabaude:
> uniqueness actually comes from ml extracted data from face being hashed.

@EmrePiconbello that doesn't sound like something you can do in a zk proof though, or that can be trusted by a smart contract and used in a permissionless way.

I'm looking at your architecture diagram https://www.mermaidchart.com/raw/4861da3a-9370-417a-b3fb-861f585f086f?theme=light&version=v0.1&format=svg

There's no detail about the role Mina or zkApps play in it - just two steps "create proof of user" and "validate proofs".
What is the content of those proofs?

It seems that proofs are created by the Pass3 backend, with the input of a face scan which was confirmed by the facetec SDK to be a unique human. The facetec SDK is able to tell you about uniqueness by asking their database about previous face scans. See [facetec.com](https://www.facetec.com/):

> For ongoing user authentication, FaceTec’s 3D face matching compares a new Liveness-proven 3D FaceMap with the user’s previously-stored 3D FaceMap

Obviously there's no way to put this interaction with facetec inside a zk proof. You can prove pure computations, but not network calls or lookups in a traditional db

* so, what concretely prevents me to create the same "proof" for 100 of my identities, by just inventing "unique hashes" and without running my face through the facetec software?
* is Pass3 a trusted actor in this system? can only they create unique identities, and do we have to trust them to not create duplicate identities?

EmrePiconbello:
@mitschabaude we are open to any solution and like to pivot if we can have a conviction on a path there.

> @EmrePiconbello thanks for your insights and research. It seems to me that the best solution, in a first iteration, might be to restrict the unique proof of personhood to people who have such a unique government id available from a signed document.
 
Extensive research needs to be done to single out the weak ones if this approach is utilised, which would limit the user base drastically. For some of the “unique” ones, there is an essential algorithm involved to generate, such as one digit being specifically one number according to the gender and such, hence it reduces the entropy drastically and opens the possibility of brute forcing.

> That doesn't sound like something you can do in a zk proof though, or that can be trusted by a smart contract and used in a permissionless way.

> There's no detail about the role Mina or zkApps play in it - just two steps "create proof of user" and "validate proofs". What is the content of those proofs?

The aim here is keeping the shared and stored data at minimal. For low risk authentication, utilising recursive proof with simple data verification(such as some basic info gathered via NFC scan) is the main part where Mina comes to the picture. Yet, due to the uncertainty of our current design in terms of public and private fields on chain due to having no certain standardisation, further examination is required to ensure the correct way of implementing the flow with least possibility of forge, as well as with least public information.

> Proofs are created by the Pass3 backend, with the input of a face scan which was confirmed by the facetec SDK to be a unique human. It seems that the facetec SDK is able to tell you about uniqueness by asking their database about previous face scans. See facetec.com:

Proofs are generated by the user side. We have the software, and we have full control. In the proposed approach, we match three elements from the user: passport data from NFC, passport data from OCR, and matching face data with the face obtained from OCR and NFC. Here, the SDK does the work on the client side. The server verifies the liveness and returns a positive or negative result. The proof is generated according to this response on the client side.

> > For ongoing user authentication, FaceTec’s 3D face matching compares a new Liveness-proven 3D FaceMap with the user’s previously-stored 3D FaceMap

> Obviously there's no way to put this interaction with facetec inside a zk proof. You can prove pure computations, but not network calls or lookups in a traditional db

Yes, unfortunately, the machine learning model is very resource intensive, making it unsuitable for running on a phone. We are not planning to utilise continuous face scan authentication. The current plan is to store the hash result from the scan on a data availability solution that we can utilise.

> * so, what concretely prevents me from creating the same "proof" for 100 of my identities, by just inventing "unique hashes" and without running my face through the facetec software?

It's not just about the face, as I shared above; all the methods need to match. You can't invent unique hashes without performing these steps. Additionally, there are only a few organisations conducting audits/benchmarks on these solutions, and they measure FRR (false rejection rate) and FAR (false acceptance rate). According to these metrics, they rank among the top performers, with rates as low as 0.0000008% in some cases.

We cannot tamper with the software SDK (I believe we can utilise the SDK inside proof) and server-side function simultaneously. Therefore, we cannot push a random hash from the server side as long as the SDK is utilised on the frontend, which will be public.

The plan involves storing the hashes on the chain, but due to limitations with Mina, we are still uncertain about how to approach this. We want the hash storage to be public if it cannot be verified with zkApp. We are still considering many options, such as verifying age without uniqueness or verifying specific data aspects, since it's customizable.

> * is Pass3 a trusted actor in this system? can only they create unique identities, and do we have to trust them to not create duplicate identities?

Pass3 is our branding; we utilise the technology from FaceTec. In every existing solution, you need to place trust in one of the many providers around the globe. They all undergo audits and benchmarks from organisations like NIST.gov, iBeta.com, etc., and comply with many standards. The issue is that you always have to give your data to a third party for compliance. For example, https://www.jumio.com/kyx/ has a direct plugin for AML screening DB provider, which performs the checks. However, for these checks, they need to share your data, or after KYC, they need to retain your data for 5-10 years, depending on regional laws.

There are only a few solutions that allow customers to build a custom solution. That's why we are not offering any compliance and legal aspects; we are simply providing verification. For each verification, we require all the steps again (because we do not plan to store any identifiable data). However, after completing a full ID enrollment, something like a signature from a private key combined with proof generated from NFC data could be sufficient to authenticate low-risk cases. Our long-term plan is to run it in a hardened confidential VM environment, where it is completely isolated. Additionally, we will strive to make it as open as possible or have 3rd party audits and certifications.

The proposal is kept simple and concise because when it comes to these details, it quickly becomes out of scope (due to the delivery timeline of three months) and complicated. Also, some decisions about the flow will be made based on the viability of zkApp and ProtoKit. Since we are not 100% familiar with the platform we are building, we do not want to create detailed flows at the start and end up delivering something completely different due to limitations.

In summary, our goal is to deliver the most private and unique ID solution we can on the platform, solving ID solutions interoperability issue. As a final note, as outlined in the proposal, many governments are currently exploring digital ID solutions, with some testing or migrating towards them. These digital IDs are mostly developed with privacy concerns in mind. Since they are directly from governments, they are also compliant. In the very long term, what we envision is that instead of KYC providers like Onfido, Sumsub, and many others collecting vast amounts of user data and relying on these solutions for compliance, they would store proof of digital IDs.

mitschabaude:
> We cannot tamper with the software SDK (I believe we can utilise the SDK inside proof)

You can't use the facetec SDK inside a proof

couragexchange (approved):

PR #12, opened 2024-01-15
RFC 0009: Wallet API - Storing & Requesting Objects

teddyjfpender:
Here is RFC-0009! It's about extending the wallet provider API to allow for storing objects, like identity credentials and proof objects.

# 🤔  Why This Is Important
A wallet's ability to interact with certain zkApps may depend on data it knows and stores, this can include credentials. In those scenarios, it is necessary for a wallet to receive credentials from issuers and provide them when requested.

# 💭 Seeking Your Input
Your expertise, feedback, and perspective are needed to ensure all requirements for a standardised wallet provider and any implementation of it!

### DIFF
--- a/RFCs/rfc-0009-wallet-provider-store-request-api.md
+++ b/RFCs/rfc-0009-wallet-provider-store-request-api.md
@@ -1,0 +1,247 @@
+# RFC-0009: Wallet-zkApp Interaction - Storing & Requesting Objects
+
+- **Intent**: Introduces an extension to the wallet provider API for storing and requesting objects like proofs and credentials.
+- **Submitted by**: Theodore Pender (Github: @teddyjfpender, email: theodore.pender@minaprotocol.com, Twitter/X: @franklyteddy)
+- **Submitted on**: Monday, January 15, 2024
+
+## Abstract
+
+This proposal details a mechanism for enhancing the interaction between zero-knowledge applications (zkApps) and wallets in the Mina ecosystem. The main objective is to allow wallets to use and store data objects that can be used when interacting with zkApps. Storing data objects provided to the wallet by zkApps and providing data objects to zkApps that are stored or created in the wallet is necessary given Mina’s ability to perform client-side proving as well as Mina's zkApps requiring provable arguments in both contract and provable program methods. It is important to note that the specification outlined in this document represents one variation of data sharing that is not censorship resistant, and does not ensure availability guarantees from the provider.
+
+## Motivation
+
+A robust integration between wallets and zkApps is essential for Mina's evolving ecosystem. It is necessary to create standards that capitalize on Mina's unique features such as client-side proving and recursive proving thereby potentially simplifying the experience for numerous protocol users. This is especially important when zkApps require arguments to either smart contracts or provable programs that do not exist on-chain nor are queryable, examples include interacting with applications that require proof objects from prior interactions, as well as applications that require verifiable credential proofs as arguments to smart contracts and provable programs. Extending the wallet Mina provider API for facilitating interactions with zkApps that have these or similar requirements can improve interaction between wallets and zkApps as well as allow zkApp developers to explore new idea-spaces for zkApp design and implementation.
+
+## Specification
+In the Web3 application ecosystem, which includes decentralized applications (dapps) and zero-knowledge applications (zkApps), it's a norm for key management software, known as "wallets," to make their API accessible through a JavaScript object in a web page—often termed "the Provider."
+
+However, inconsistencies have arisen in Provider implementations across wallets. This proposal seeks to standardize a Mina Provider API to ensure continued wallet interoperability with zkApps. This specification assumes wallets already implement the minimal, event-driven APIs equivalent to EIP-1193 for Mina; this specification introduces new RPC methods and message event types.
+
+Although Providers have historically been presented as `window.mina` in browsers, this convention isn't mandated in the specification.
+
+### API
+Wallets should have a provider API parallel to EIP-1193 but tailored for Mina. The goal of this API extension is to build on top of pre-existing standards. 
+
+#### Key Features of the Proposed API
+
+- **Agnosticity to Application Types**: Whether an application operates on-chain, off-chain, or a combination of the two, this API remains consistent and reliable. 
+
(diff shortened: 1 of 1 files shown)


RFCs/rfc-0009-wallet-provider-store-request-api.md
```diff
@@ -0,0 +1,247 @@
...
+    - **User Consent for Storing Objects**: Just as access is restricted, storage of any objects by the wallet requires the user's green light. It ensures that users are aware of what's being stored in their wallets, enhancing transparency.
+
+#### Primary Functionalities:
+The proposed API should empower a Mina wallet to:
+- **Receive Objects**: Accept and store objects, whether they're verifiable credentials, proofs, or any other data.
+- **Request Objects**: Initiate requests to retrieve stored objects, facilitating interactions like transaction validations or data computations.
+
+### Scenario: A Wallet Receiving Objects to Store
+#### `mina_setState`
+Providers within Mina's domain could introduce a new RPC method: `mina_setState`. When called, a user interface might emerge, allowing users to choose whether to store objects from a zkApp. This method returns a Promise: resolving with a boolean or, if storing is unavailable, rejecting with an Error. A zkApp MUST provide sufficient information in the request to allow the wallet to store and index the object appropriately.
```
es92, line 41: What does it mean to "store objects from a zkApps"? I'm imagining that some zkApps would have public state - and others would store hashes of private state, and the private state is the thing that this is storing? Or something else?

RFCs/rfc-0009-wallet-provider-store-request-api.md
```diff
@@ -0,0 +1,247 @@
...
+
+ [2] `RESOLVE`
+The Promise returned when calling the `mina_setState` RPC method MUST be resolved with a boolean. This must emit an event `message` indicating that the request to store the object has been approved and the promise has been resolved with a boolean.
+
+ [3] `REJECT`
+The Promise returned when calling the `mina_setState` RPC method MUST be rejected with an informative Error if no storage is available for any reason. This must emit an event `message` indicating that the request to store the object has been rejected, and the promise is rejected with an error.
+
+A wallet MUST adequately detail to the user what the zkApp is requesting the wallet to store. If approved the wallet MUST appropriately store the object for future use. The zkApp COULD provide in the params appropriate information for how this object is indexed if the zkApp will require the information later.
+
+#### Example initialization - Request objects to be stored in the wallet
```
es92, line 70: Is it possible this is not a pattern we would want to allow? It seems if the information is private, we might not want to release the information to the website at all (vs releasing proofs to the website ie the attestation standard API)

Or are there cases where there is private data, that we do want websites to be able to request directly, and can't / don't want to just share a proof?

RFCs/rfc-0009-wallet-provider-store-request-api.md
teddyjfpender: A zkApp can produce a well-formed credential that has a Kimchi proof in it, in that case, the zkApp would be the `issuer` of the credential and can allow the wallet to store that object for future use like in a smart contract method argument. 

If one were to assume that these objects could be fetched, or rather requested from a zkApp, this response can be stored much like how transaction history or current account information is stored in the wallet. The objects the wallet stores don't have to be private information but they can be.

A wallet storing credentials they can use in smart contract or provable-program methods is more about how to improve the interaction design between wallets and zkApps. If it is beneficial for a wallet to hold those credentials then they can use this API as a way of enabling that, there of course exist other designs that don't require the wallet to hold credentials but that design pattern is outside the scope of this RFC :)

RFCs/rfc-0009-wallet-provider-store-request-api.md
teddyjfpender: This pattern would be for public, not sensitive information. Instead of a KYC credential that might contain personally identifiable information, a wallet might store a zkOracle attestation that makes a claim about some Web 2 data, in that case the user doesn't need to attest to that data again, it can just provide the original attestation proof it knows of.

In any case, a wallet should communicate to the user what a zkApp is requesting to ensure the user is completely aware of what data they are sharing. Data stored in a wallet should be encrypted at rest to prevent unauthorised access to it, and partitioned well enough to separate concerns and not accidentally provide private data.

Pfed-prog: From my understanding the data could be stored as:

app_id  - data 

potentially on a merkle tree with app_id as index, 
this could enable even easier data provision into an app that waits for a merkle witness and data

RFCs/rfc-0009-wallet-provider-store-request-api.md
lvshaoping007: But how can users ensure these credentials are beneficial ?

RFCs/rfc-0009-wallet-provider-store-request-api.md
lvshaoping007: How does zkApp know whether the wallet has the information he needs? And how does the wallet know what information zkApp needs?

es92: > This pattern would be for public, not sensitive information. Instead of a KYC credential that might contain personally identifiable information, a wallet might store a zkOracle attestation that makes a claim about some Web 2 data, in that case the user doesn't need to attest to that data again, it can just provide the original attestation proof it knows of.
> 
> In any case, a wallet should communicate to the user what a zkApp is requesting to ensure the user is completely aware of what data they are sharing. Data stored in a wallet should be encrypted at rest to prevent unauthorised access to it, and partitioned well enough to separate concerns and not accidentally provide private data.

What would "the original attestation proof it knows of" look like? Is this caching proofs from the attestation API? Or something more general? Think I"m missing what other kinds of things this would want to expose.

RFCs/rfc-0009-wallet-provider-store-request-api.md
teddyjfpender: > But how can users ensure these credentials are beneficial ?

That is at up to the wallet user, the requested storing of those credentials must detail what the credential or object being stored is to the wallet user for them to determine the benefit. The real benefit of storing credentials in the wallet is that it gives the wallet user the ability to utilise them as and when they need/desire, without relying on central parties.

TyrellCorp2020:
Moving to draft as will need further review before progressing

PR #13, opened 2024-01-15
RFC 0008: Standard JavaScript Mina Wallet Provider API 

teddyjfpender:
Here is RFC-0008! It's about introducing a standard JavaScript Mina Wallet Provider API to the ecosystem.

# 🤔  Why This Is Important
There exists no standard way for wallets and zkApps to interact. Standardising this API means zkApps have to spend less time integrating against many wallets, allowing any wallet to interact with any zkApp.

# 💭 Seeking Your Input
Your expertise, feedback, and perspective are needed to ensure that all requirements for a standardised wallet provider and any implementation are met!

### DIFF
--- a/RFCs/rfc-0008-wallet-provider-api.md
+++ b/RFCs/rfc-0008-wallet-provider-api.md
@@ -175,7 +175,7 @@ Provider.request({
 ```
 
 #### mina_chainInformation
-Returns both the chain ID and an informative name of the chain
+Returns both the chain ID and an informative name of the chain. The name should follow the [CAIP-2 specification for Mina](https://github.com/ChainAgnostic/namespaces/blob/main/mina/caip2.md#test-cases).
 
 ##### Parameters
 None
--- a/RFCs/rfc-0008-wallet-provider-api.md
+++ b/RFCs/rfc-0008-wallet-provider-api.md
@@ -174,6 +174,20 @@ Provider.request({
 }): Promise<string>;
 ```
 
+#### mina_chainInformation
+Returns both the chain ID and an informative name of the chain
+
+##### Parameters
+None
+
+##### Returns
+```typescript
+type AddChainArgs = {
+   readonly url: string;
+   readonly name: string;
+};
+```
+
 #### mina_sign
 Returns a signed message.
 
@@ -209,12 +223,9 @@ Returns a signed message.
 
 ```typescript
 type SignedFieldsData = {
-    publicKey: string;
-    data: (string | number)[];
-    signature: {
-        field: string;
-        scalar: string;
-    };
+  data: (string | number)[];
+  publicKey:string;
+  signature: string;
 };
 
 Provider.request({
@@ -254,7 +265,7 @@ type TransactionReceipt = {
 
 Provider.request({
     method: 'mina_sendTransaction',
-    params: { signedTransaction: signedTx, transactionBody: txBody, transactionType: 'payment'}
+    params: { unsignedTransaction: unsignedTx, transactionBody: txBody, transactionType: 'payment'}
 }): Promise<TransactionReceipt>
 ```
 #### mina_getBalance
@@ -297,6 +308,40 @@ Provider.request({
 }): Promise<Nullifier>
 ```
 
+#### mina_switchChain
+
+##### Parameters
+```typescript
+type SwitchChainArgs = {
+   readonly chainId: string;
+};
+```
+
+##### Returns
+```typescript
+type AddChainArgs = {
+   readonly url: string;
+   readonly name: string;
+};
+```
+
+#### mina_addChain
+
+##### Parameters
+```typescript
+type AddChainArgs = {
+   readonly url: string;
+   readonly name: string;
+};
+```
+
+##### Returns
+```typescript
+type AddChainArgs = {
+   readonly url: string;
+   readonly name: string;
+};
+```
 
 ## Appendix I: Consumer-Facing API Documentation
 
--- a/RFCs/rfc-0008-wallet-provider-api.md
+++ b/RFCs/rfc-0008-wallet-provider-api.md
@@ -1,8 +1,8 @@
 # RFC-0008: Mina Provider JavaScript API
 
-## Summary
-
-A JavaScript Mina Provider API for consistency across clients and applications.
+- **Intent**: Introduces a first, extensible standard JavaScript Mina Provider API for consistency across clients and applications.
+- **Submitted by**: Theodore Pender (Github: @teddyjfpender, email: theodore.pender@minaprotocol.com, Twitter/X: @franklyteddy)
+- **Submitted on**: Monday, January 15, 2024
 
 ## Abstract
 
--- a/RFCs/rfc-0008-wallet-provider-api.md
+++ b/RFCs/rfc-0008-wallet-provider-api.md
@@ -1,0 +1,387 @@
+# RFC-0008: Mina Provider JavaScript API
+
+## Summary
+
+A JavaScript Mina Provider API for consistency across clients and applications.
+
+## Abstract
+
+This RFC looks to formalizes a Mina Provider API to promote wallet interoperability within the Mina ecosystem. The API is designed to be minimal, event-driven, and agnostic of transport and RPC protocols.
+
+## Specification
+
+### Definitions
+
+- **Provider**: A JavaScript object that provides access to Mina by means of a Client.
+- **Client**: An endpoint that receives Remote Procedure Call (RPC) requests from the Provider and returns their results.
+- **Wallet**: An application that manages private keys and acts as a middleware between the Provider and the Client.
+
+### Connectivity
+
+The Provider is "connected" when it can service RPC requests to at least one chain.
+
+### API
+
+#### request
+
+The `request` method is a transport- and protocol-agnostic wrapper function for Remote Procedure Calls (RPCs).
+
+```typescript
+interface RequestArguments {
+  readonly method: string;
+  readonly params?: readonly unknown[] | object;
+}
+
+Provider.request(args: RequestArguments): Promise<unknown>;
+```
+
+### Supported RPC Methods
+
+A "supported RPC method" is any RPC method that may be called via the Provider.
+
+All supported RPC methods MUST be identified by unique strings. The unique identifiers must be consistent across all wallet providers.
+
+Providers MAY support whatever RPC methods required to fulfill their purpose, standardized or otherwise.
+
+If an RPC method defined in a finalized RFC (or MIP) is not supported, it SHOULD be rejected with a `4200` error per the Provider Errors section below, or an appropriate error per the RPC method's specification.
+
+### RPC Errors
+
+```typescript
+interface ProviderRpcError extends Error {
+  message: string;
+  code: number;
+  data?: unknown;
+}
+```
+
+#### Error Standards
+
+`ProviderRpcError` codes and messages should follow these conventions:ß◊
+- `4001`: User Rejected Request
+- `4100`: Unauthorized
+- `4200`: Unsupported Method
+- `4900`: Disconnected
+- `4901`: Chain Disconnected
+
+### Events
+
+The Provider **MUST** implement the following event handling methods:
+- `on` 
+- `removeListener` methods. 
+
+These methods MUST be implemented per the Node.js EventEmitter API.
+
+
+It also should handle the `mina_message` and `connect` events.
+
+#### mina_message
+
+This event is for arbitrary notifications. When emitted, the `mina_message` event must be with an object argument:
+
+```typescript
+interface ProviderMessage {
+  readonly type: string;
+  readonly data: unknown;
+}
+```
+
+#### connect
+
+The connect event must be emitted with an object:
+
+```typescript
+interface ProviderConnectInfo {
+  readonly chainId: string;
+}
+```
+
+#### disconnect
+
+The Provider is said to be "disconnected" when it cannot service RPC requests to any chain at all.
+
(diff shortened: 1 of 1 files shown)


RFCs/rfc-0008-wallet-provider-api.md
```diff
@@ -0,0 +1,387 @@
...
+```
+
+#### mina_signTransaction
+Returns a signed transaction.
+
+##### Parameters
+`transaction`: [Required] a well-formed transaction body.
+
+##### Returns
+A signed transaction body.
```
lvshaoping007, line 233: Why does zkApp need a signed transaction body ?

RFCs/rfc-0008-wallet-provider-api.md
```diff
@@ -0,0 +1,387 @@
...
+
+```typescript
+interface ProviderConnectInfo {
+  readonly chainId: string;
+}
+
+Provider.on('connect', listener: (connectInfo: ProviderConnectInfo) => void): Provider;
+```
+
+The event emits an object with a hexadecimal string `chainId` per the `mina_chainId` Mina GraphQL query result, and other properties as determined by the Provider.
```
lvshaoping007, line 343: There seems to be no fixed chainId now, and the provider needs to define an enum to enumerate different chainId.

RFCs/rfc-0008-wallet-provider-api.md
```diff
@@ -0,0 +1,387 @@
...
+```
+
+#### mina_chainId
+Returns the chain ID of the current network.
+
+##### Parameters
+None
+
+##### Returns
+A hexadecimal of the current chain ID.
```
lvshaoping007, line 169: Is it possible to add a chain name here, for chain with the same chainId but different GQL URLs ?

RFCs/rfc-0008-wallet-provider-api.md
```diff
@@ -0,0 +1,387 @@
...
+}): Promise<SignedMessage>;
+```
+
+#### mina_signFields
+Returns a signed message.
+
+##### Parameters
+`fields`: [Required] an array of numbers or strings
+
+##### Returns
```
lvshaoping007, line 208: The current return type of `[mina-signer].signFields` is the following, which is different from `[mina-signer].signMessage`
```typescript
type SignedFieldsData = {
  data: (string | number)[];
  publicKey:string;
  signature: string;
}
```

RFCs/rfc-0008-wallet-provider-api.md
```diff
@@ -0,0 +1,387 @@
...
+- `transactionType`: [Required] the type of the transaction (e.g. `payment`, `delegation`, `zkapp`)
+
+```typescript
+type TransactionReceipt = {
+    hash: string
+};
+
+Provider.request({
+    method: 'mina_sendTransaction',
+    params: { signedTransaction: signedTx, transactionBody: txBody, transactionType: 'payment'}
```
lvshaoping007, line 257: Perhaps `unsignedTransaction` and `unsignedTx` should be used here ?

RFCs/rfc-0008-wallet-provider-api.md
lvshaoping007: And should we provide a method to `verifyFields` and `verifyMessage`?

RFCs/rfc-0008-wallet-provider-api.md
```diff
@@ -0,0 +1,387 @@
...
+    };
+};
+
+Provider.request({
+    method: 'mina_createNullifier',
+    params: { message: [1n] }
+}): Promise<Nullifier>
+```
+
+
```
lvshaoping007, line 300: Here you need some APIs about the chain, such as requesting to add or switch chains
For example: 
```typescript
mina_switchChain
// Request parameters:
type SwitchChainArgs = {
   readonly chainId: string;
};
mina_addChain:
// Request parameters:
type AddChainArgs = {
   readonly url: string;
   readonly name: string;
};

// Both of return parameters
type ChainInfoArgs = {
   chainId: string;
   name: string;
};
```

RFCs/rfc-0008-wallet-provider-api.md
teddyjfpender: This can also be sign and submit, but not all scenarios may require a transaction to be submitted immediately after being signed.

RFCs/rfc-0008-wallet-provider-api.md
teddyjfpender: An application like a wallet must be aware of the information for the chain it is "connected" to, how it chooses to do that is an implementation detail.

RFCs/rfc-0008-wallet-provider-api.md
teddyjfpender: It can do, but is that to make it easier for a zkApp to understand information about the chain? The `mina_chainId` method serves to _only_ provide the current network's chain id -- if there is another method that you think is important to have such as `mina_chainInformation` what would you believe would be an appropriate return object/type?

RFCs/rfc-0008-wallet-provider-api.md
teddyjfpender: > And should we provide a method to verify signed fields and message ?

I'm happy to add in verification methods! I wonder what scenario they would be useful to a zkApp however, verification of a message or signed field might be an internal wallet feature rather than for a zkApp to request the verification of a wallet. A malicious wallet could of course choose not to provide the correct verification result!

RFCs/rfc-0008-wallet-provider-api.md
teddyjfpender: Agreed! 🦾

RFCs/rfc-0008-wallet-provider-api.md
teddyjfpender: What wallets can use to provide more information is something like [this](https://github.com/ChainAgnostic/namespaces/pull/91). In that case, the wallet can provide that standardised `networkID` (e.g. `mina:mainnet`)

RFCs/rfc-0008-wallet-provider-api.md
```diff
@@ -0,0 +1,387 @@
...
+If the Provider becomes disconnected from all chains, the Provider MUST emit the event named `disconnect` with value error: ProviderRpcError, per the interfaced defined in the RPC Errors section. The value of the error's code property MUST follow the status codes for CloseEvent.
+
+#### chainChanged
+
+If the chain the Provider is connected to changes, the Provider MUST emit the event named `chainChanged` with value `chainId: string`, specifying the integer ID of the new chain as a hexadecimal string, per the `mina_chainId` RPC method.
+
+#### accountsChanged
+
+If the accounts available to the Provider change, the Provider MUST emit the event named `accountsChanged` with value `accounts: string[]`, containing the account addresses per the `mina_accounts` RPC method.
```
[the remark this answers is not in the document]
teddyjfpender: Switching to an account that is not available to the provider would mean the user has to reconnect, upon reconnecting that account becomes available and therefore the wallet must emit the `accountsChanged` event.

However if the user does not re-connect then no event should be triggered because there has not been a change to the accounts available.

Thoughts?

RFCs/rfc-0008-wallet-provider-api.md
lvshaoping007: Yes, because some users need to add different proxies for the same network, which can be distinguished by name
This is the structure we return to zkApp now, mainly used to distinguish the network actually used by zkApp
```ts
type ChainInfoArgs = {
  chainId: string;
  name: string;
};
```

RFCs/rfc-0008-wallet-provider-api.md
lvshaoping007: In fact, I just assumed that this scenario existed, but I haven't come across a place where I need to use this. It’s just that we tend to expose the `mina-signer` method to give zkApp greater freedom.

RFCs/rfc-0008-wallet-provider-api.md
lvshaoping007: Yes, we are currently doing this. We define the network ID inside the Auro Wallet to distinguish different networks, because the network ID returned by the node will now change.
```ts
const netType = {
  Mainnet: "mainnet",
  Devnet: "devnet",
  Berkeley: "berkeley",
  Testworld2: "testworld2"
}
```

RFCs/rfc-0008-wallet-provider-api.md
```diff
@@ -0,0 +1,387 @@
...
+
+It also should handle the `mina_message` and `connect` events.
+
+#### mina_message
+
+This event is for arbitrary notifications. When emitted, the `mina_message` event must be with an object argument:
+
+```typescript
+interface ProviderMessage {
+  readonly type: string;
```
lvshaoping007, line 84: What is the role of type here? Is it to distinguish message types?
We want to add a type, or specification, here. When zkApp passes specific parameters, the message data will be displayed in a format, making the signed information content more readable.

just like this 
image

RFCs/rfc-0008-wallet-provider-api.md
```diff
@@ -0,0 +1,387 @@
...
+  readonly params?: readonly unknown[] | object;
+}
+
+Provider.request(args: RequestArguments): Promise<unknown>;
+```
+
+The returned Promise resolves with the method's result or rejects with a [`ProviderRpcError`](#errors). For example:
+
+```javascript
+Provider.request({ method: 'mina_accounts' })
```
lvshaoping007, line 319: After zkApp request, when the current wallet account is not connected to zkApp, there are two methods:
- return silently
- A popup will pop up requesting to link the account

Because some zkApps hope to get zkApp requests silently, so as to avoid forced pop-ups as soon as zkApp is opened.

RFCs/rfc-0008-wallet-provider-api.md
```diff
@@ -0,0 +1,387 @@
...
+
+The Provider is intended to pass messages between a Mina Client and a Mina application. It is not responsible for private key or account management; it merely processes RPC messages and emits events. Consequently, account security and user privacy need to be implemented in middlewares between the Provider and its Mina Client. In practice, we call these middleware applications "Wallets," and they usually manage the user's private keys and accounts. The Provider can be thought of as an extension of the Wallet, exposed in an untrusted environment, under the control of some third party (e.g. a website).
+
+#### Handling Adversarial Behavior
+
+Since it is a JavaScript object, consumers can generally perform arbitrary operations on the Provider, and all its properties can be read or overwritten. Therefore, it is best to treat the Provider object as though it is controlled by an adversary. It is paramount that the Provider implementer protects the user, Wallet, and Client by ensuring that:
+
+- The Provider does not contain any private user data.
+- The Provider and Wallet programs are isolated from each other.
+- The Wallet and/or Client rate-limit requests from the Provider.
```
lvshaoping007, line 141: `rate-limit` Does this refer to the frequency with which zkApp requests the wallet?

RFCs/rfc-0008-wallet-provider-api.md
lvshaoping007: What should zkApp do if it needs to recommend memo or fees? Does transactionBody directly put the processed content?

RFCs/rfc-0008-wallet-provider-api.md
```diff
@@ -0,0 +1,387 @@
...
+```typescript
+Provider.on('accountsChanged', listener: (accounts: string[]) => void): Provider;
+```
+
+The event emits with `accounts`, an array of account addresses, per the `mina_accounts` RPC method.
+
+#### message
+
+The Provider emits `message` to communicate arbitrary messages to the consumer.
+Messages may include JSON-RPC notifications, GraphQL subscriptions, and/or any other event as defined by the Provider.
```
lvshaoping007, line 378: We should add some json-rpc that zkApp can actively request

NorbertBodziony:
IMO we should also consider autodetection of wallets in this RFC. API of networks like Solana are based on https://github.com/wallet-standard/wallet-standard. This skips a lot of issues related to supply chain attack. ETH recently also released EIP addressing this issue https://eips.ethereum.org/EIPS/eip-6963.

Happy to make MVP after wallet interface gets finalized.

RFCs/rfc-0008-wallet-provider-api.md
kantp: @lvshaoping007 I'm curious to understand why one would introduce different names for the same network. What is the purpose of having different proxies for the same network?

RFCs/rfc-0008-wallet-provider-api.md
kantp: Yes, the `type` can be used to have different message types, defined by the provider.

RFCs/rfc-0008-wallet-provider-api.md
kantp: Yes, limiting the frequency with which requests will be accepted, as a protection mechanism.

RFCs/rfc-0008-wallet-provider-api.md
kantp: I've added the `mina_chainInformation` method.

RFCs/rfc-0008-wallet-provider-api.md
kantp: I've updated the `SignedFieldsData` type.

RFCs/rfc-0008-wallet-provider-api.md
kantp: Added!

kantp (dismissed):
Looks good to me. Thanks for all the comments and suggestions @lvshaoping007!

chad11111 (dismissed):

halsaphi (approved):

PR #14, opened 2024-01-15
RFC 0010: Wallet Attestation API

teddyjfpender:
Here is RFC-0010! It's about extending the wallet provider API to enable the wallet to provide attestations of private data, like credentials, it knows about.

# 🤔  Why This Is Important
Much like in [RFC-0009](https://github.com/MinaFoundation/Core-Grants/pull/12), a wallet's ability to interact with certain zkApps may depend on data it knows and stores; this can include credentials or privately proving knowledge about certain data. In the latter scenario, it is necessary for a wallet to receive a request for attestation that binds a wallet to a specific scenario to prove it knows about certain data within its application context and not in the browser context.

# 💭 Seeking Your Input
Your expertise, feedback, and perspective are needed to make sure all requirements for a standardised wallet provider and any implementation of it!

### DIFF
--- a/RFCs/rfc-0010-wallet-provider-attestation-api.md
+++ b/RFCs/rfc-0010-wallet-provider-attestation-api.md
@@ -12,6 +12,14 @@ The motivation behind extending the Mina wallet provider API stems from the evol
 
 ## Specification
 
+### Requirements
+
+* Private data can’t leave the wallet sandbox
+* No remote code execution in the wallet
+* Work for multiple projects without requiring wallet changes / coordination, so it works over the many different projects and any new private data attestation projects that come up
+* Should tie into nullifiers (to make sure there is an application specific salt)
+
+
 ### API
 Wallets should have a provider API parallel to EIP-1193 but tailored for Mina. The goal of this API extension is to build on top of pre-existing standards. 
 
@@ -24,7 +32,10 @@ Wallets should have a provider API parallel to EIP-1193 but tailored for Mina. T
 
 #### Primary Functionalities:
 The proposed API should empower a Mina wallet to:
+- **Store Credential**: Store credentials to be used by an attestation program.
 - **Request Attestation**: Initiate requests to receive an attestation proof of a claim.
+- **Request Nullifier Key**: Construct a nullifier key, along with an attestation proof about that nullifier key
+
 
 ### Scenario: Requesting Attestation From a Wallet
 #### `mina_requestAttestation`
@@ -64,10 +75,10 @@ When a requester seeks attestation from a wallet, they must provide contextual b
 #### `mina_requestAttestation` Binding Parameters
 
 #### `program`
-A unique identifier or hash representing the provable program the wallet should run internally for attestation. This might correspond to the verification key of the provable program or its hash.
+A unique identifier or hash representing the provable program the wallet should run internally for attestation. Programs will be added via the standards process, and will initially include one program, "flat-credential-attestation".
 
 ### `clientData`
-An adaptable data structure designed to bind the wallet to contextually relevant information. The content and structure of `clientData` are dynamic, tailored to the specific scenario or attestation requirement at hand. It can encapsulate a range of data, from verifiable credentials to custom datasets, based on what the wallet is expected to know when attesting. Given the myriad potential use cases, the shape and constituents of clientData can vary widely. Requesters should structure it based on prior knowledge of the data the wallet is acquainted with, ensuring effective communication and validation during the attestation process.
+An adaptable data structure designed to bind the wallet to contextually relevant information. The content and structure of `clientData` are dynamic, tailored to the specific program.
 
 #### `publicInputs`
 A key-value map detailing the provable-program's method signature for open inputs. Each key-value pair describes a specific input and the value to use in the provable-program; further binding the wallet.
@@ -83,26 +94,11 @@ try {
     // Define the binding for the attestation
     const binding = {
         // The program could be constrained by a unique identifier for the program like a verification key
-        program: "ProvableProgramForRoyaltyClaim",
-        clientData: {
-            "type": ["VerifiableCredential", "ArtOwnershipCredential"],
-            "credentialData": {
-                "issuer": "did:mina:B62sj...owg",
-                "credentialSubject": {
-                    "id": "did:mina:B62rf...kf9",
-                }
-            }
-        },
+        program: "flat-credential-attestation",
+        clientData: {},
         // Define the required public inputs
         publicInputs: {
-            "claim": {
-                "property": "credentialData.credentialSubject.royaltyPercentage",
-                "operation": "eq",
-                "value": "5"
-            },
-            "nonce": {
-                "value": "1"
-            }
+            "verifier": constructVerifierForRoyaltyClaim(),
         },
     };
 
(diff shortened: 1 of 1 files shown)


RFCs/rfc-0010-wallet-provider-attestation-api.md
```diff
@@ -0,0 +1,252 @@
...
+```ts
+import { Mina } from "o1js";
+
+const mina = window.mina;
+
+try {
+    // Define the binding for the attestation
+    const binding = {
+        // The program could be constrained by a unique identifier for the program like a verification key
+        program: "ProvableProgramForRoyaltyClaim",
```
es92, line 86: How does this program selection work? Is it something already hardcoded into the wallet, or does it make it in another way?

If hardcoded, maybe we should consider something that operates over json records? Like instead of a string being able to specify proofs over statements on records? (Like I think we talked about at an early version of this, getting a field and proving its type and basic things about it like for strings checking if it matches another string, for numbers inequalities, etc)

RFCs/rfc-0010-wallet-provider-attestation-api.md
```diff
@@ -0,0 +1,252 @@
...
+
+    // Request attestation using the binding. Wallet will match and verify internally against stored credentials.
+    const attestationProof = await mina.request({
+        method: 'mina_requestAttestation',
+        params: binding
+    });
+
+    // Construct a transaction using the received attestation proof
+    let transaction = await Mina.transaction(() => {
+        new RoyaltiesZkApp(zkappAddress).claim(attestationProof);
```
es92, line 117: Would claim be a function on the zkApp, that takes in a Proof type? (like [here](https://docs.minaprotocol.com/zkapps/o1js/recursion)?)

RFCs/rfc-0010-wallet-provider-attestation-api.md
```diff
@@ -0,0 +1,252 @@
...
+        CONTINUE zkProvableProgram
+    IF user rejects
+        REJECT[3] attestation
+        STOP interaction
+IF provider is undefined
+    STOP interaction
+```
+
+[1] REQUEST
+The requester MUST request attestation by calling the `mina_requestAttestation` RPC method on the provider exposed at `window.mina`. Calling this method MUST trigger a user interface allowing the user to approve or reject the attestation request. This method MUST return a Promise that is resolved with the desired proof or rejected if the attestation cannot be produced.
```
es92, line 54: Should the request include information on what information will be revealed to the website? Maybe if so, programs should include data about what information they are proving for the UI?

RFCs/rfc-0010-wallet-provider-attestation-api.md
teddyjfpender: The `binding` is part of the request the application makes to the wallet. So the `program` field tells the wallet which provable program to use, this naively assumes the wallet knows what program that is and it's background script can fetch/compile the program such that it can execute the program with the defined inputs (public or private).

This certainly needs inputs from wallet providers to understand what is possible to implement. Holistically, I see these provable programs as little zkApps that are modules in the wallet that can be added & removed.

RFCs/rfc-0010-wallet-provider-attestation-api.md
teddyjfpender: Yes! It would be a method on the contract, or possibly another provable-program!

RFCs/rfc-0010-wallet-provider-attestation-api.md
teddyjfpender: Indeed -- that's an important component, a wallet user _must_ be able to understand precisely what they are attesting to and with what inputs. I'll add a section on this.

RFCs/rfc-0010-wallet-provider-attestation-api.md
es92: I would propose then,we do something that operates over records, so that people can add new attestations without needing to re-coordinate and get updates from wallet providers? I think we should add that to this spec as well (lmk, if needed I can find the older doc I think we had on it too somewhere...)

es92: Thinking about this a bit more, I think that one option could be:

- Have attestations be json records (with extra types for different crypto primitives; and maybe flat json records)
- Make a DSL that operates over json records, and can be safely sent to the wallet to be executed in the wallet's context without security risks
- Have the DSL able to provide the following:
    - Binary statement over whether an attestation passes the check performed by the DSL, to compute the proof
        - this can also check which attestations would pass the DSL, to ask the user which to use if they have multiple attestations that would pass (or let them know if they have none)
    - Detail so the wallet can notify the user what information the DSL is asking for from the attestation (eg this will check [ Age > 20, proof of unique identity wrt nullifier X, etc ]

The DSL could be like is done for o1js, where its just a typescript library;

Maybe something like 
```
(attestation) => {
   const verifier = new AttestationVerifier(attestation);
   return verifier.check_equal("field1", "value")
              .and(verifier.check_signature("field2", verifier.get("data_field"), "publicKey"))
              .and(verifier.check_gt("field3", 20));
}
```

I think ideally we'd want an ecosystem of attestations, so hopefully this could help make it easier to produce / consume attestations permissionlessly. Curious on any thoughts!

teddyjfpender: The `bindings` should satisfy those requirements -- they're about constraining the wallet to attest to something specific:
- Which provable-program to use
- Who the expected issuer of the private data is
- What the expected logic gate is (e.g. `age > 20`)

Since a wallet at it's core is just a key management system, it is not impossible to imagine an attestation management system that can sit next to the KMS or integrate into it. Much like a generic `api.sign(params)` can be one API of that key management system `api.prove(params)` could be too. That `params` argument is the `bindings` and should serve the purpose as might a `api.sign(TransactionBody)` might. Wallet users can already attest to statements in the browser context like constructing transactions (they are the spender/fee payer of that transaction), only to pass to the wallet for signing/submission, the key difference with this attestation API is the wallet is asked to use the `bindings` as context for producing the required proof within its own application.

In that sense, the attestation API _could_ be used as a way for a zkApp to give context to the wallet about how to construct a transaction, sign it, and prove it before submitting to the network. I drew a little diagram to show the shift in responsibilities:
Screenshot 2024-01-18 at 10 32 25

For this attestation context it is about giving the wallet context on how to construct arguments for a provable-program method and prove it.

es92: What does bindings look like exactly in this model? I'm still a little confused exactly where the proof if computed, and how exactly it is computed?

I was thinking it probably needs to be computed in the wallet, so that the attestation data doesn't leave the wallet to the unsafe webpage, but then the wallet needs some specific way of knowing how to compute the proof.

es92: In particular, we say above the wallet would be told "Which provable-program to use" - to me that sounds like wallets would have a fixed set of programs they can use - which seems limiting if we want new attestation producers and consumers to come along and build new attestations and consume them in their own way.

teddyjfpender: > What does bindings look like exactly in this model?

The bindings are [defined here](https://github.com/MinaFoundation/Core-Grants/blob/rfc-0010-wallet-provider-private-attestation-api/RFCs/rfc-0010-wallet-provider-attestation-api.md#mina_requestattestation-binding-parameters). 

> where the proof if computed, and how exactly it is computed?

The bindings define what program and what inputs are expected via `ClientData` and `publicInputs`, proving should happen in the wallet application (or any secure enclave). `bindings` is the data structure that this RFC needs the most feedback on as it defines what `provable-program` and what inputs the wallet should use. For example, a wallet should be bound (hence the name `bindings`) to prove a specific `Issuer` created a credential for their `PublicKey`; they shouldn't be able to prove that _any_ `Issuer` created that credential.

> "Which provable-program to use" to me that sounds like wallets would have a fixed set of programs they can use 

It should be read as "which provable-program to use for this attestation". There is no limit on the set of provable-programs a wallet can know about. As long as a wallet can know (a.) where to obtain the specific program, (b.) compile it, and (c.) cache the program to prevent always re-compiling when attesting, there isn't a limitation (modulo browser/service worker constraints).

> I think we would still need something like the above DSL, so that we can allow dynamic proofs over new

The attestations created can be stored as credentials in the wallet's storage for re-use, but this somewhat can conflict with the considerations section where attestations may require a TTL.

es92: Ah I see - so can the bindings can define a series of operations constituting a claim (which could be constructed via DSL, or could be statically defined as done here)?

I think then I'm not understanding the role of the provable-program? Like what does changing that done?

I think I'm proposing then, in this new context, to
- Only provide information on the claim, as a statement over the fields of attestations (and not have provable-program? Unless I'm probably missing why its necessary)
- Maybe use a DSL to construct claims to make it easier? Though probably not necessary, especially for a first version.

es92: Just to share context here from speaking offline, we met and confirmed it does make sense to have "provable-programs" - but that provable-programs should be probably added via the standards process, and this standard should include a first provable program, “RecordCredentialAttestation” or something of the like, that operates with boolean statements over structured json-like attestation objects. If I have time later, I’ll make a PR on this one with what that could look like.

PR #24, opened 2024-02-13
RFC 0012 - Access Control Standards

ChrisCates:
The goal of this RFC is to standardize access control over various functions in Mina Smart Contracts. As of now, there are three core flows:

1. Single Signature

2. Multi Signature

3. Role Based Access Control

The Access Control Standard should take inspiration from Open Zeppelin formatting and standards, and, leverage Typescript decorations in order to make access control among functions easy to implement.

### DIFF
--- a/RFCs/rfc-0012-access-control.md
+++ b/RFCs/rfc-0012-access-control.md
@@ -1,0 +1,244 @@
+# RFC-0012: Access Control
+
+- **Intent**: Access Control Standards for Single and Multi Signature Access
+- **Submitted by**: Chris Cates, <hello@chriscates.ca>, @ChrisCates on Discord
+- **Submitted on**: February 13th, 2024
+
+## Abstract
+
+The goal of this RFC is to standardize access control over various functions in Mina Smart Contracts. As of now, there are three core flows:
+
+1. Single Signature
+
+2. Multi Signature
+
+3. Role Based Access Control
+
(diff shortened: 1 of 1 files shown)


teddyjfpender:
This is an awesome idea!

I think there can be a couple more requirements/investigation areas that can be added to this RFC to create an artefact that looks at investigating those deemed most relevant and impactful.

1. Provable Programs as Signature Schemes
    - Provable programs can be used recursively to obtain signatures and verifiable if all a threshold of specified signatures have been received. This can work as a standard provable-program template in `o1js` but simply require different sets of signers for verification.
    - Of course consideration would have to be made for attacks (e.g. replay attacks) so programs and contracts would have to go hand-in-hand to ensure that the level of security is guaranteed.

2. Contracts threshold signature verification
   - Without a provable program being passed around between signers (`1` or `n`) and just using a contract method, I think it would be fantastic to look at using the `o1js` [ECDSA](https://docs.minaprotocol.com/zkapps/o1js/ecdsa) features to natively verify, particularly to verify threshold signatures schemes.

ChrisCates:
@teddyjfpender, thank you for the feedback. Adding these two use cases as an extension to an npm module definitely makes a ton of sense to me.

If possible, could you share some Github repositories demonstrating the usage in O1JS? I will update and fine tune the RFC after reviewing some example source code.

Thank you!

PR #25, opened 2024-02-15
RFC-0013: Soulboud Token Standard

kantp:

Soulbound tokens are an interesting concept with a variety of potential applications. This rfc aspires to be a first step towards establishing a standard for how to create and interact with them on Mina.

# 🤔 Why This Is Important

Creating a standard for soublound tokens will make it easier for developers to create them, and, perhaps more importantly, ease integration for third party contracts and wallets.

# 💭 Seeking Your Input

Your expertise, feedback, and perspective are needed to make sure we end up with a suitable standard.

### DIFF
--- a/RFCs/rfc-0011-soulbound-tokens.md
+++ b/RFCs/rfc-0011-soulbound-tokens.md
@@ -1,0 +1,153 @@
+# RFC-0011: Soulboud Token Standard
+
+- **Intent**: Create a standard for Soulbound Tokens on Mina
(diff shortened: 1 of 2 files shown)


kantp:
Moving this to draft for now, as we want to unify this with the other token standards that are being worked on.

mazito:
I think this is really needed sooner than later. 
We could use it for issuing Socialcap credentials.
Currently we are issuing them with our own contract, but would be really good to be bounded to a standard.

PR #32, opened 2024-04-24
RFC 0014 DA landscape discussion

yunus433:


RFCs/rfc-0014-data-availability.md
```diff
@@ -0,0 +1,182 @@
...
+
+Moreover, even though sequencers do not have the power to change a TX they receive in any way, they decide on the ordering of TXs happening in the same settlement block (thus the name “sequencer”). In financial applications, this creates a problem of arbitrage: Sequencers may put their TXs before some significant buy / sell order to gain considerable amounts of money. In a way, sequencers can see the near future and act accordingly.
+
+The most efficient way of increasing the liveness of a rollup is to include a DA layer in the process. If users submit their TXs into a DA layer (through their own light client) and the sequencer is only responsible for submitting them on Mina, then the problem of censorship or arbitrage disappears. As said before, almost all zkApps on Mina are rollups, so integrating a DA layer natively into Mina is very important for zkApp developers.
```
kantp, line 71: I don't understand why the problem of censorship and arbitrage disappears in this setting. Does the sequencer not choose the order in which they submit transactions from the DA layer to Mina in this setting?

RFCs/rfc-0014-data-availability.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Objectives
+
+This RFC is mainly to start a discussion around possible ways of integrating a DA into Mina and get the developer community involvement in the process. There are for sure a lot of unspoken advantages and bottlenecks missing in this RFC, and it is precisely the motivation to raise some discussion points around these issues.
```
Trivo25, line 75: I agree with this section, but I think you should provide a few more starting points.
- Where to integrate DA? Is this a rollup-only thing, at the core of the Mina protocol or at the zkApp level?
- What are the trade offs with different approaches, especially when it comes to devX/ergonomic and efficiency?

michal0mina:
Closing for now as there is no clear deliverable. Lets keep it as discussion reference 

It will be revisited after upcoming improvements to o1js offchain storage.

PR #36, opened 2024-07-02
RFC-0010: ZK Passport

chad11111:
migrated from private repo

### DIFF
--- a/RFCs/rfc-0010-zk-passport.md
+++ b/RFCs/rfc-0010-zk-passport.md
@@ -1,0 +1,147 @@
+## RFP: ZKPassport
+
+### Abstract
+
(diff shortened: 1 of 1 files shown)


RFCs/rfc-0010-zk-passport.md
```diff
@@ -0,0 +1,147 @@
...
+### Specification
+
+Generating a proof will involve multiple steps:
+
+#### Register Circuit
+
+The register circuit is used for the following:
+
+- Verify the signature of the passport
+- Verify that the public key which signed the passport is part of the registry Merkle tree (a check of the Merkle roots will be performed on-chain)
```
es92, line 32: What is the registry merle tree in this case? I'm not sure there is one, unfortunately (or at least not a public one)

RFCs/rfc-0010-zk-passport.md
```diff
@@ -0,0 +1,147 @@
+## RFP: ZKPassport
```
es92, line 1: Can you migrate over relevant content from the prior RFC? I think there's a lot of relevant information in there about requirements, etc https://github.com/MinaFoundation/Core-Grants/pull/11

